PR #1, opened 2016-02-26
Improve parallel pack generation

phmccarty:
Because pack generation is a time-consuming task, running this task in
parallel is preferable.

However, if we are generating two or more packs with the same "from"
version at the same time, the multiple jobs will use the same pack
staging directory, leading to corruption.

My solution for now is to use a staging dir with a name derived from
both "from" and "to" versions so that there are no collisions.

A long-term fix will be to implement some kind of locking mechanism to
prevent _identical_ packs from being generated simultaneously, but for
now, using more unique staging directories will improve the situation.

### DIFF
--- a/src/pack.c
+++ b/src/pack.c
@@ -39,28 +39,31 @@
 
 #include "swupd.h"
 
-static void empty_pack_stage(int full, int version, char *module)
+static void empty_pack_stage(int full, int from_version, int to_version, char *module)
 {
 	char *cmd;
 	char *path;
 	int ret;
 
 	// clean any stale data (eg: re-run after a failure)
-	string_or_die(&cmd, "rm -rf %s/%s/%i/", packstage_dir, module, version);
+	string_or_die(&cmd, "rm -rf %s/%s/%i_to_%i/", packstage_dir, module,
+		      from_version, to_version);
 	ret = system(cmd);
 	if (ret) {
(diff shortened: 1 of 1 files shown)


src/pack.c
```diff
@@ -39,28 +39,31 @@
...
-	string_or_die(&cmd, "rm -rf %s/%s/%i/", packstage_dir, module, version);
+	string_or_die(&cmd, "rm -rf %s/%s/%i_to_%i/", packstage_dir, module,
+		      from_version, to_version);
 	ret = system(cmd);
 	if (ret) {
-		fprintf(stderr, "Failed to clean %s/%s/%i\n",
-			packstage_dir, module, version);
+		fprintf(stderr, "Failed to clean %s/%s/%i_to_%i\n",
+			packstage_dir, module, from_version, to_version);
 		exit(EXIT_FAILURE);
```
bryteise, line 55: Probably could free(cmd) just in case someday we return instead of exit here.

ikeydoherty, line 55: At this point I'd suggest looking into an `autofree()` system like I use in some of my stuff

phmccarty, line 55: @bryteise Thanks, I will add the free().

@ikeydoherty That would be a nice general cleanup, since string_or_die() is used all over the place. I'll take a look at your implementation.

phmccarty:
I rebased this change and pushed to master.

PR #2, opened 2016-02-29
Always use xattrs when calculating file hashes

rojkov:
Currently swupd-client includes xattrs to hash sums of all files.
Therefore in order to avoid relaxing security and to prevent
`swupd verify` from reporting hash mismatches for updated files
it's better to include xattrs to hashes for all files.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -793,7 +793,8 @@ bool compute_hash_with_xattrs(const char *filename)
 		return true;
 	}
 
-	return false;
+	/* Currently swupd-client includes xattrs to hash sums of all files */
+	return true;
 }
 
 /* Returns 0 == success, -1 == failure */


bradTpeters:
This is a problematic change. While I agree it is a good idea, it is going to force massive delta pack creation and file-system wide updates. What is the download requirement?

I suggest you use our test system to test this change, and collect some stats on the update scope.  How long does a client update take? How many bytes need be downloaded? 

Depending on answer, we might want to discuss options, such as rolling this into the next format bump, or perhaps even using our pre-update scripts support to manually set xattrs prior to the update which mandates this hash calculation (thus no longer requiring a mass download)

rojkov:
This change is not going to force massive delta pack creation and causes zero impact on update size if CLR doesn't really use extended attribute. And I believe CLR doesn't use xattrs beyond /boot, because `swupd verify` would report massive hash mismatches even before first update otherwise.

Look at how `hmac_compute_key()` calculates hash sums: if a file doesn't have xattrs set (like in CLR's case) then `xattrs_blob_len == 0` no matter what `compute_hash_with_xattrs()` returns.

Unfortunately I don't have access to the test system, but my local tests with a Yocto-based distro don't show any increase in delta download size. Without the change `swupd verify` reports become useless on IMA-, SMACK- or SELinux-enabled systems.

bradTpeters:
It's not the files without xattr's used which concerns me, its the files using them. The scale of the delta pack creation and update size could be zero, or it could be massive. After poking around a bit, it seems xattr's are not broadly used, so I'm inclined to believe the update would be rather small. But it sure would be nice to have hard data.

bradTpeters:
Dmitry has done several tests to determine the size of the potential impact of this change. Given that it seems only the systemd journal currently uses extended attributes, the impact is very small.

+1

phmccarty:
This change looks fine to me.

Since Clear Linux packages do not install files with xattrs, it does not impact Clear Linux right now. If Clear Linux does ship files with xattrs in the future, this change will be beneficial.

And if another distro using swupd-server does ship files with xattrs, swupd-client is currently enabled to support hash calculation with xattrs by default (hardcoded), so no action needed for swupd-client.

phmccarty:
Rebased and pushed to master. Thanks!

PR #3, opened 2016-03-09
Add compatibility with libarchive's bsdtar command

rojkov:
Since GNU tar fails to extract files with xattrs preserved when
Integrity Measurement Architecture (IMA) is enabled some vendors
may choose to install libarchive-based tar (bsdtar) on their embedded
devices, so the swupd server needs to be able to create archives
in its format.

This patch adds one compile-time options --enable-bsdtar that is used
to enable/disable GNU tar specific options. Also it harmonizes
the command strings to be compatible with both GNU tar and bsdtar.
Particularly it
- changes --exclude pattern from `'%s'/*` to more explicit `'%s/?*'` because
  bsdtar's pattern matching is greedier than in tar: it uses tcsh's
  globbing where "*" can be anything including the null string and the
  original pattern would include the directory itself;
- OS file names are escaped with leading ./ to avoid collisions with
  file names starting with @ which has special meaning in bsdtar.

PS. I understand the team may not want to support two flavors of tar, but as users of the tool we have a use case for that and have to maintain our patches ourselves. And since our development policy requires us to submit PRs upstream for the patches we maintain I submit this PR.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -29,6 +29,15 @@ AC_ARG_ENABLE(
 )
 
 AC_ARG_ENABLE(
+	[bsdtar],
+	AS_HELP_STRING([--enable-bsdtar],[Use alternative bsdtar command (uses tar by default)]),
+	AS_IF([test "x$enable_bsdtar" = "xyes"],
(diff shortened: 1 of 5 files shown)


phmccarty:
I pushed this change to master. Thanks!

PR #4, opened 2016-03-15
Remove unneeded script

tmarcu:
This functionality is now performed by the chroot builder.


phmccarty:
basic_creator.sh uses this script currently, so it cannot be removed unless the dependency is removed for the basic creator...

PR #5, opened 2016-03-17
Includes

bryteise:
This patch set adds support for including submanifests on the server and requires a format bump before being released.

### DIFF
--- a/src/analyze_fs.c
+++ b/src/analyze_fs.c
@@ -23,19 +23,19 @@
  */
 
 #define _GNU_SOURCE
-#include <stdio.h>
-#include <stdlib.h>
-#include <unistd.h>
-#include <string.h>
 #include <assert.h>
+#include <dirent.h>
 #include <errno.h>
 #include <glib.h>
-#include <sys/types.h>
-#include <dirent.h>
-#include <sys/stat.h>
-#include <sys/mman.h>
-#include <openssl/hmac.h>
 #include <linux/limits.h>
+#include <openssl/hmac.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
 
 #include "swupd.h"
 #include "xattrs.h"
--- a/src/analyze_fs.c
+++ b/src/analyze_fs.c
@@ -402,6 +402,57 @@ struct manifest *full_manifest_from_directory(int version)
 	return manifest;
 }
 
+/* Read includes from $manifest-includes file */
+GList *get_sub_manifest_includes(char *component, int version)
+{
+	FILE *infile;
+	GList *includes = NULL;
+	char *c;
+	char *conf;
+	char *filename;
+	char *included;
+	char line[8192];
+
+	conf = config_output_dir();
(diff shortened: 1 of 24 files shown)


src/fullfiles.c
```diff
@@ -92,8 +91,7 @@ static void create_fullfile(struct file *file)
 			assert(0);
 		}
 
-		string_or_die(&tarcommand, TAR_COMMAND " -C %s " TAR_PERM_ATTR_ARGS " -cf - --exclude='%s/?*' './%s' 2> /dev/null | "
-			      TAR_COMMAND " -C %s " TAR_PERM_ATTR_ARGS " -xf - 2> /dev/null",
+		string_or_die(&tarcommand, TAR_COMMAND " -C %s " TAR_PERM_ATTR_ARGS " -cf - --exclude='%s/?*' './%s' 2> /dev/null | " TAR_COMMAND " -C %s " TAR_PERM_ATTR_ARGS " -xf - 2> /dev/null",
```
phmccarty, line 94: I don't really like that clang-format joins lines like this. Maybe because it doesn't run the preprocessor, it gets confused that TAR_COMMAND is really a string constant?

src/manifest.c
```diff
@@ -554,6 +564,13 @@ void subtract_manifests(struct manifest *m1, struct manifest *m2)
 		}
 		list2 = g_list_next(list2);
 	}
+
+	/* now recurse into the included manifests and subtract those as well */
+	while (includes) {
+		mi = includes->data;
+		includes = g_list_next(includes);
+		subtract_manifests(m1, mi, includes);
```
phmccarty, line 580: After the recursive calls return, "includes" will again be non-NULL, so I'm seeing lot's of duplicate subtractions happening. Can we avoid the duplicates?

src/analyze_fs.c
```diff
@@ -402,6 +402,57 @@ struct manifest *full_manifest_from_directory(int version)
...
+		if (c) {
+			*c = 0;
+		}
+		if (strlen(line) == 0) {
+			break;
+		}
+		included = strdup(line);
+		includes = g_list_prepend(includes, included);
+	}
+	includes = g_list_sort(includes, file_sort_filename);
```
phmccarty, line 451: file_sort_filename() expects the list items to be "struct file". So, I'm seeing a segfault when processing a Clear bundle that has includes. Not sure why the functional test is passing, haven't looked closely at that yet.

src/manifest.c
phmccarty, line 580: I'm also not seeing all included bundles being subtracted.

For example, "devtools-extras" includes "R-extras", which includes "R-basic". However, I'm only seeing "R-extras" subtracted.

phmccarty, line 580: Last thing I've noticed so far: the subtraction for the "openstack-all-in-one" bundle seems to be in an infinite recursion loop. Since the number of subtractions happening is quite large and includes lots of duplicates, I haven't determined why this is happening yet...

src/fullfiles.c
bryteise, line 94: Yea, that was annoying though I wonder if it is just a line length issue?

bryteise:
Okay, quite a few fixups here, replaced with #7.

PR #6, opened 2016-03-18
Fix manifest subtraction logic

phmccarty:
In subtract_manifests(), if the two files under consideration have the
same filename but should not be subtracted, a file is skipped in the m2
manifest because there is no "continue" statement at the end of the
first "if" block.

This results in potentially many files not getting subtracted when they
ought to be.

To clarify the logic, avoid using "continue", since it is not needed.
Convert to using one "if" block instead.


bryteise:
+1

amshinde:
+1

PR #7, opened 2016-03-21
Includesv2

bryteise:
Fixed the following from #5 :
- Free bundle includes file stream.
- Duplicate bundle subtraction
- Nested bundle includes


phmccarty:
+1 to the full patch set. Since this is a breaking change for format3, I can merge this series after the format4 bump.

PR #8, opened 2016-03-22
Speed up pack creation

phmccarty:
A big chunk of time during pack creation is spent in untar'ing of
fullfiles tar files.  However, we also (most likely) have this exact
same content already in the image/ directory.

This patch makes the pack creator skip the untar if its possible to link
the file directly, and falls back to the existing behavior if the link
fails


bryteise:
+1 though we'll want to pull this into a seperate function at some point.

PR #9, opened 2016-03-23
Capture logs for every binary

phmccarty:
To capture all information that is logged, change init_log() to make
every log file name unique, and add a call to swupd_make_fullfiles to
enable logging.


bryteise:
+1

PR #10, opened 2016-03-30
Format handling improvements

phmccarty:
Over time, manifests accumulate lots of deleted files which bloat the manifest. At the time of a format increment, the deleted file entries in manifests prior to the current version can be pruned; the first patch implements this change.

Also, remove the hardcoding of the format "3" in the swupd_create_update binary. Instead, always require the -F option. In the future, I imagine that the default value could be read from a config file, and -F would override, but for now just require -F.

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -79,7 +79,8 @@ bool set_state_dir(char *dir)
 bool init_globals(void)
 {
 	if (format == -1) {
-		format = SWUPD_DEFAULT_FORMAT;
+		printf("Error: Missing format parameter. Please specify a format with -F.\n");
+		return false;
 	}
 
 	if (!init_state_globals()) {
--- a/src/globals.c
+++ b/src/globals.c
@@ -33,7 +33,7 @@
 
 int newversion = -1;
 int minversion = 0;
-int format = -1;
+unsigned long long int format = 0;
 bool enable_signing = false;
 
 char *state_dir = NULL;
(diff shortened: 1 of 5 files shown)


src/globals.c
```diff
@@ -41,20 +41,17 @@ char *packstage_dir = NULL;
...
 {
-	int version;
+	int user_format;
 
-	// expect a positive integer
+	// format string shall be a positive integer
 	errno = 0;
-	version = strtoull(userinput, NULL, 10);
-	if ((errno < 0) || (version <= 0)) {
+	user_format = strtoull(userinput, NULL, 10);
```
bryteise, line 50: Setting an int to an unsigned long long.

src/globals.c
```diff
@@ -41,20 +41,17 @@ char *packstage_dir = NULL;
...
+	if ((errno < 0) || (user_format <= 0)) {
```
bryteise: == 0 as per above conversion.

bryteise:
I think you probably just want to use atoi for the conversion instead of strtoull. Otherwise this looks good to me.

src/globals.c
phmccarty, line 50: I will fix the variable type.

src/globals.c
phmccarty: Will fix

phmccarty:
Since atoi() doesn't indicate any error conditions with the conversion, I would prefer to continue using strtoull().

phmccarty:
Added some changes from the review. I will squash them before the merge.

phmccarty:
Merged.

PR #11, opened 2016-03-30
Fix manifests not detecting include changes

bryteise:
When deciding if a new manifest should be generated for a new version,
the included manifests were not being used to detect a manifest
change. This caused manifests that had new includes: lines in the new
version to not be generated resulting in files from the included
manifest to be missed when adding/updating that bundle.

The fix was to compare includes: lines from the current and previous
manifest versions for differences and generate new manifests when that
was the case. In order to compare manifests duplicate includes: lines
are not allowed and previously os-core could be added multiple times to
a manifest because it was always added. This change also added detection
for os-core before automatically adding it to a bundle.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -877,6 +877,44 @@ void sort_manifest_by_version(struct manifest *manifest)
 	}
 }
 
+bool manifest_includes(struct manifest *manifest, char *component)
+{
+        GList *includes = g_list_first(manifest->includes);
+
+        while (includes) {
+                if (strcmp(((struct manifest *)includes->data)->component,
+                           component) == 0) {
+                        return true;
+                }
+                includes = g_list_next(includes);
+        }
+
+        return false;
+}
+
+bool changed_includes(struct manifest *old, struct manifest *new)
+{
+        GList *includes_old = g_list_first(old->includes);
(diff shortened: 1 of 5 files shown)


src/manifest.c
```diff
@@ -877,6 +877,44 @@ void sort_manifest_by_version(struct manifest *manifest)
...
+        if (g_list_length(includes_old) != g_list_length(includes_new)) {
+                return true;
+        }
+
+        while (includes_old && includes_new) {
+                char *bundle_old = ((struct manifest *)includes_old->data)->component;
+                char *bundle_new = ((struct manifest *)includes_new->data)->component;
+                if (strcmp(bundle_old, bundle_new) != 0) {
+                        return true;
+                }
```
phmccarty, line 909: Should we assume the includes lists are sorted before calling this function? Comparing the bundle names like this assumes that.

bryteise, line 909: Yes and they are sorted, see https://github.com/clearlinux/swupd-server/blob/master/src/analyze_fs.c#L452. Deserves a comment though.

phmccarty, line 909: Okay, thanks for clarifying. I would like to see the code comment added.

Otherwise, +1

bryteise:
Pushed ammended patch to master.

PR #12, opened 2016-04-14
Include file

bryteise:
This patch set improves the performance of fully resolving include hierarchies and adds an Manifest.MoM.includes file that lists each bundle's fully resolved include list.


src/manifest.c
```diff
@@ -88,6 +88,21 @@ int file_sort_filename(gconstpointer a, gconstpointer b)
...
+	struct manifest *A, *B;
+	A = (struct manifest *)a;
+	B = (struct manifest *)b;
+
+	ret = strcmp(A->component, B->component);
+	if (ret) {
+		return ret;
+	}
+
+	return 0;
```
phmccarty, line 103: For simplicity, this can just return "ret", no need for the conditional.

src/manifest.c
```diff
@@ -727,6 +742,97 @@ static int write_manifest_signature(struct manifest *manifest, const char *suffi
...
+	char *base = NULL;
+	char *conf = config_output_dir();
+	char *dir = NULL;
+	char *filename = NULL;
+	int ret = -1;
+	unsigned int count = 0;
+
+	/* ignore non MoM manifests */
+	if (strcmp(manifest->component, "MoM")) {
+                ret = 0;
```
phmccarty, line 760: weird indentation here

src/manifest.c
```diff
@@ -727,6 +742,97 @@ static int write_manifest_signature(struct manifest *manifest, const char *suffi
...
+		goto exit;
+	}
+
+	if (conf == NULL) {
+		assert(0);
```
phmccarty, line 765: I would prefer assert(conf != NULL)

src/manifest.c
```diff
@@ -727,6 +742,97 @@ static int write_manifest_signature(struct manifest *manifest, const char *suffi
...
+	}
+	string_or_die(&filename, "%s/%i/Manifest.%s.includes", conf,
+		      manifest->version, manifest->component);
```
phmccarty, line 768: Because the file body of this new file is completely different than the other manifests, should we call it a manifest? I would be okay calling it an "includes-list" or similar.

src/manifest.c
```diff
@@ -727,6 +742,97 @@ static int write_manifest_signature(struct manifest *manifest, const char *suffi
...
+
+	base = strdup(filename);
+	if (base == NULL) {
+		assert(0);
```
phmccarty, line 772: Similarly here, I would prefer to do assert(base != NULL)

src/manifest.c
```diff
@@ -815,33 +921,41 @@ static int write_manifest_plain(struct manifest *manifest)
...
 {
 	char *conf = config_output_dir();
+	char *files = NULL;
+	char *append = NULL;
 	char *tarcmd = NULL;
 	int ret = -1;
 
 	if (conf == NULL) {
 		assert(0);
 	}
```
phmccarty, line 931: It looks like there's already a convention... :-)  I'm okay with changing the assert convention separately.

phmccarty:
+1 on the excess recursion commit.

src/manifest.c
bryteise, line 931: Yea, would be nice to change all the asserts to not be condition + assert(0). I'll leave that for future work though since it'd be good to do all of it in one run.

bryteise:
Fixes in #13

PR #13, opened 2016-04-19
Include filev2

bryteise:
Fixes from #12:
- manifest_sort_compare just returns ret
- update include metadata filename to includes-list
- incorrect indentation


phmccarty:
I think we should defer this change, and instead add the recursive resolving for 'bundle-remove' to make it work correctly with bundle includes.

phmccarty:
Instead of using a slightly different manifest format for the includes file, I think we should use JSON instead. The current manifests will continue to use the custom tab-separated format with the in-tree parser.

phmccarty:
Closing this request, since @tmarcu will be working on a new implementation using JSON for the include file.

PR #14, opened 2016-04-22
Change delta's name from FROM-TO-HASH2 to FROM-TO-HASH1-HASH2 format.

jrguzman-intel:
There is an issue when two different files in a newer release have same hash:
This is, when swupd updates a file by applying a delta, it takes the HASH2 to
know the file where delta must be applyed. If files are different in current
release (before updating), there must be 2 deltas, one for each file but as
the two files have same hash in new release delta's name are the same:
FROM-TO-HASH2 and it is when issue arises due to just the last delta file is
kept when swupd server creates files and packs. So when swupd client tries to
apply the delta to one of the file that does not corresponds it will fail and
generates an error. At the first look it will seem like delta file is corrupted
however the issue is that delta file was created for another file.
To solve this issue we include the hash for the original file in the delta's
name so that swupd client can take the correct delta and apply it:
FROM-TO-HASH1-HASH2.

Warning: A corresponding patch to swupd client must be applied in order to
sync both sides and understand the new delta's name format.

->
Refer to https://github.com/jrguzman-intel/swupd-client/commit/49175ce96aae78515d70a6cd6e7aed1291247f98
-<

### DIFF
--- a/include/swupd.h
+++ b/include/swupd.h
@@ -238,7 +238,7 @@ extern void type_change_detection(struct manifest *manifest);
 
 extern void rename_detection(struct manifest *manifest, int last_change, GList *last_versions_list);
 extern void link_renames(GList *newfiles, struct manifest *from_manifest);
-extern void __create_delta(struct file *file, int from_version);
+extern void __create_delta(struct file *file, int from_version, char *from_hash);
 
 extern void account_delta_hit(void);
 extern void account_delta_miss(void);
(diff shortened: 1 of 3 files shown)


bryteise:
Looks good but like the client I'd like to see some tests for this.

jrguzman-intel:
@bryteise It was tested by generating files and then feed swupd-client with new deltas. Also fullfiles were deleted to make sure swupd-client uses deltas to update files. This was two step testing one without changes in swupd-client, so an error was occurred in the update and it was expected and then with the proper patch in client side and update was successful.

tmarcu:
Merged into master with supplementary conflict fixes.

PR #15, opened 2016-04-29
Report full paths for files containing blacklisted characters

phmccarty:
To better track which files are excluded from manifests because they
contain blacklisted characters, the full path is more useful to log
rather than the file basename.


tmarcu:
+1 from me

PR #16, opened 2016-05-03
[clearlinux/swupd-server] Replace system calls by local system_argv functions.

jrguzman-intel:
There are new functions that replace system() calls: system_argv(), system_argv_fd and system_argv_pipe().
They are based on execvp() function and they can manage any argument including filenames with characters
that could be interpreted as meta-characters in a regular system() function.
So no escape for the arguments needed and is less prone to errors.

### DIFF
--- a/include/swupd.h
+++ b/include/swupd.h
@@ -252,7 +252,7 @@ extern void dump_file_info(struct file *file);
 extern void string_or_die(char **strp, const char *fmt, ...);
 extern void print_elapsed_time(const char *step, struct timeval *previous_time, struct timeval *current_time);
 extern int system_argv(char *const argv[]);
-extern int system_argv_fd(char *const argv[], int stdin, int stdout, int stderr);
+extern int system_argv_fd(char *const argv[], int newstdin, int newstdout, int newstderr);
 extern int system_argv_pipe(char *const argvp1[], int stdinp1, int stderrp1,
 							char *const argvp2[], int stdoutp2, int stderrp2);
 
--- a/include/swupd.h
+++ b/include/swupd.h
@@ -19,17 +19,21 @@
 #if SWUPD_WITH_BSDTAR
 #define TAR_COMMAND "bsdtar"
 #define TAR_XATTR_ARGS ""
+#define TAR_XATTR_ARGS_STRLIST
 #define TAR_WARN_ARGS ""
 #else
 #define TAR_COMMAND "tar"
 #define TAR_XATTR_ARGS "--xattrs --xattrs-include='*'"
+#define TAR_XATTR_ARGS_STRLIST "--xattrs", "--xattrs-include='*'",
 #define TAR_WARN_ARGS "--warning=no-timestamp"
 #endif
 
 #if SWUPD_WITH_SELINUX
 #define TAR_PERM_ATTR_ARGS "--preserve-permissions --selinux " TAR_XATTR_ARGS
+#define TAR_PERM_ATTR_ARGS_STRLIST TAR_XATTR_ARGS_STRLIST "--preserve-permissions", "--selinux"
 #else
 #define TAR_PERM_ATTR_ARGS "--preserve-permissions " TAR_XATTR_ARGS
+#define TAR_PERM_ATTR_ARGS_STRLIST TAR_XATTR_ARGS_STRLIST "--preserve-permissions"
 #endif
 
 #if SWUPD_WITH_STATELESS
@@ -248,6 +252,9 @@ extern void dump_file_info(struct file *file);
 extern void string_or_die(char **strp, const char *fmt, ...);
 extern void print_elapsed_time(const char *step, struct timeval *previous_time, struct timeval *current_time);
 extern int system_argv(char *const argv[]);
+extern int system_argv_fd(char *const argv[], int stdin, int stdout, int stderr);
+extern int system_argv_pipe(char *const argvp1[], int stdinp1, int stderrp1,
+							char *const argvp2[], int stdoutp2, int stderrp2);
 
 extern bool signature_initialize(void);
 extern void signature_terminate(void);
--- a/src/chroot.c
+++ b/src/chroot.c
@@ -34,9 +34,8 @@
 
 void chroot_create_full(int newversion)
 {
-	int ret;
 	char *group;
-	char *command;
+	char *param;
 	char *full_dir;
 
 	string_or_die(&full_dir, "%s/%i/full/", image_dir, newversion);
@@ -45,11 +44,12 @@ void chroot_create_full(int newversion)
 
 	/* start with base */
 	LOG(NULL, "Copying chroot os-core to full", "");
-	string_or_die(&command, "rsync -aAX %s/%i/os-core/ %s",
-		      image_dir, newversion, full_dir);
-	ret = system(command);
-	assert(ret == 0);
-	free(command);
+	string_or_die(&param, "%s/%i/os-core/", image_dir, newversion);
+	char *const rsynccmd[] = { "rsync", "-aAX", param, full_dir, NULL };
+	if (system_argv(rsynccmd) != 0) {
+		assert(0);
+	}
+	free(param);
 
 	/* overlay any new files from each group */
 	while (1) {
@@ -59,11 +59,12 @@ void chroot_create_full(int newversion)
 		}
 
 		LOG(NULL, "Overlaying bundle chroot onto full", "%s", group);
-		string_or_die(&command, "rsync -aAX --ignore-existing %s/%i/%s/ %s",
-			      image_dir, newversion, group, full_dir);
-		ret = system(command);
-		assert(ret == 0);
-		free(command);
+		string_or_die(&param, "%s/%i/%s/", image_dir, newversion, group);
+		char *const rsynccmd[] = { "rsync", "-aAX", "--ignore-existing", param, full_dir, NULL };
+		if (system_argv(rsynccmd) != 0) {
+			assert(0);
+		}
+		free(param);
 	}
 	free(full_dir);
 }
--- a/src/delta.c
+++ b/src/delta.c
@@ -37,8 +37,8 @@
 
 void __create_delta(struct file *file, int from_version)
 {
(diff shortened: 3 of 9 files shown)


include/swupd.h
```diff
@@ -19,17 +19,21 @@
 #if SWUPD_WITH_BSDTAR
 #define TAR_COMMAND "bsdtar"
 #define TAR_XATTR_ARGS ""
+#define TAR_XATTR_ARGS_STRLIST
```
bryteise, line 22: Does bsdtar not handle xattrs?

jrguzman-intel, line 22: @bryteise As far as I remember bsdtar manage extended attributes by default, you don't need to specify.

src/manifest.c
```diff
@@ -807,34 +807,42 @@ static int write_manifest_plain(struct manifest *manifest)
...
 	/* now, tar the thing up for efficient full file download */
 	/* and put the signature of the plain manifest into the archive, too */
 	if (enable_signing) {
-		string_or_die(&tarcmd, TAR_COMMAND " --directory=%s/%i " TAR_PERM_ATTR_ARGS " -Jcf "
-						   "%s/%i/Manifest.%s.tar Manifest.%s Manifest.%s.signed",
-			      conf, manifest->version, conf, manifest->version, manifest->component,
-			      manifest->component, manifest->component);
+		char *const tarcmd[] = { TAR_COMMAND, directory, TAR_PERM_ATTR_ARGS_STRLIST, "-Jcf",
+								 manifesttar, manifestcomp, manifestsigned, NULL};
+		if (system_argv(tarcmd) != 0) {
```
bryteise: Since you are duplicating the error handling maybe just assign the tarcmd in the enable_signing if clause and do the system_argv and error checking outside that block.

src/pack.c
```diff
@@ -491,12 +488,12 @@ static int make_final_pack(struct packdata *pack)
...
-	free(tar);
+	string_or_die(&param1, "--directory=%s/%s/%i_to_%i/", packstage_dir, pack->module, pack->from, pack->to);
+	string_or_die(&param2, "%s/%i/pack-%s-from-%i.tar", staging_dir, pack->to, pack->module, pack->from);
+	char *const tarcmd[] = { TAR_COMMAND, TAR_PERM_ATTR_ARGS_STRLIST, param1, "--numeric-owner", "-Jcf", param2, "delta", "staged", NULL };
+	ret = system_argv(tarcmd);
+	free(param1);
+	free(param2);
 	LOG(NULL, "finished tar for pack", "%s: %i to %i", pack->module, pack->from, pack->to);
 	/* FIXME: this is a hack workaround, needs diagnosed and removed */
 	if ((ret != 0) && (ret != 256)) {
```
bryteise: Does the ret check need to be updated here?

src/manifest.c
jrguzman-intel: @bryteise You're right. Will fix.

src/pack.c
jrguzman-intel: @bryteise Not sure about this. @phmccarty Did you know the history of this sentence?

phmccarty: @jrguzman-intel I'm not sure about the historical meaning of the FIXME comment, so I would leave it for now.

That said, @bryteise is correct that the ret check needs to be updated. Since system_argv() returns the WEXITSTATUS value of the waitpid() exit code, the check should be (ret != 0) && (ret != 1).

src/delta.c
```diff
@@ -110,7 +109,12 @@ void __create_delta(struct file *file, int from_version)
...
 	}
 
-	ret = system(sanitycheck);
+	string_or_die(&param1, "%s", newfile);
+	string_or_die(&param2, "%s", testnewfile);
+	char *const sanitycheck[] = { "cmp", "-s", param1, param2, NULL};
+	ret = system_argv(sanitycheck);
+	free(param1);
+	free(param2);
 	if (ret == -1 || !WIFEXITED(ret) || WEXITSTATUS(ret) == 2) {
```
phmccarty: system_argv() already calls WIFEXITED and WEXITSTATUS, so this line needs updating.

src/chroot.c
```diff
@@ -79,11 +80,12 @@ void chroot_create_full(int newversion)
...
-		ret = system(command);
-		assert(ret == 0);
-		free(command);
+		string_or_die(&param, "%s/%i/%s/", image_dir, newversion, group);
+		char *const rsynccmd[] = { "rsync", "-aAX", "--existing", full_dir, param, NULL };
+		if (system_argv(rsynccmd) != 0) {
+			assert(0);
+		}
+		free(param);
 	}
```
phmccarty: This section of code was removed in latest master. Can you rebase?

src/manifest.c
```diff
@@ -807,34 +807,38 @@ static int write_manifest_plain(struct manifest *manifest)
...
-			      manifest->component);
+		char *const tarcmd[] = { TAR_COMMAND, directory, TAR_PERM_ATTR_ARGS_STRLIST, "-Jcf",
+								 manifesttar, manifestcomp, NULL};
+		ret = system_argv(tarcmd);
 	}
-
-	if (system(tarcmd) != 0) {
+	if (ret) {
 		fprintf(stderr, "Creation of Manifest.tar failed\n");
-		goto exit;
```
phmccarty, line 866: Removing this goto means that the "exit" label is now unused. Should we keep the goto, since this is an error condition?

src/versions.c
```diff
@@ -126,35 +127,51 @@ void ensure_version_image_exists(int version)
...
 
 	conf = config_output_dir();
 	if (conf == NULL) {
 		assert(0);
 	}
 
-	string_or_die(&cmd, "echo %s > %s/%i/swupd-server-src-version", VERSION, conf, version);
-	if (system(cmd) != 0) {
+	string_or_die(&filename, "%s/%i/swupd-server-src-version", conf, version);
+	versionfd = open(filename, O_WRONLY | O_CREAT | O_TRUNC, 0664);
```
phmccarty: the mode should be 0644 instead

src/versions.c
```diff
@@ -126,35 +127,51 @@ void ensure_version_image_exists(int version)
...
+		assert(0);
+	}
+	close(versionfd);
+	free(filename);
 
-	string_or_die(&cmd, "echo %llu > %s/%i/format", format, conf, version);
-	if (system(cmd) != 0) {
+	string_or_die(&filename, "%s/%i/format", conf, version);
+	string_or_die(&strformat, "%llu", format);
+	formatfd = open(filename, O_WRONLY | O_CREAT | O_TRUNC, 0664);
```
phmccarty: Similarly here, 0644

src/versions.c
```diff
@@ -126,35 +127,51 @@ void ensure_version_image_exists(int version)
...
 
-	string_or_die(&cmd, "echo %s > %s/%i/swupd-server-src-version", VERSION, conf, version);
-	if (system(cmd) != 0) {
+	string_or_die(&filename, "%s/%i/swupd-server-src-version", conf, version);
+	versionfd = open(filename, O_WRONLY | O_CREAT | O_TRUNC, 0664);
+	if (versionfd == -1) {
 		assert(0);
 	}
-	free(cmd);
+	if (write(versionfd, VERSION, strlen(VERSION)) != strlen(VERSION)) {
```
phmccarty: The call to write() can be interrupted, or it may not write all data at first (unlikely in this case, but better to be safe). Or you could use fopen/fwrite instead.

src/versions.c
```diff
@@ -126,35 +127,51 @@ void ensure_version_image_exists(int version)
...
-	string_or_die(&cmd, "echo %llu > %s/%i/format", format, conf, version);
-	if (system(cmd) != 0) {
+	string_or_die(&filename, "%s/%i/format", conf, version);
+	string_or_die(&strformat, "%llu", format);
+	formatfd = open(filename, O_WRONLY | O_CREAT | O_TRUNC, 0664);
+	if (formatfd == -1) {
 		assert(0);
 	}
-	free(cmd);
+	if (write(formatfd, strformat, strlen(strformat)) != (ssize_t)strlen(strformat)) {
```
phmccarty: Same comment as above for this write() call.

src/helpers.c
```diff
@@ -181,9 +181,93 @@ int system_argv(char *const argv[])
...
+		return -1;
 	}
+	statusp1 = system_argv_fd(argvp1, stdinp1, pipefd[1], stderrp1);
+	close(pipefd[1]);
+	statusp2 = system_argv_fd(argvp2, pipefd[0], stdoutp2, stderrp2);
+	close(pipefd[0]);
+
+	/* Returns the status of the failed process if any
+       If both processes failed returns the status of first one */
+	return statusp1 ? statusp1 : statusp2;
```
phmccarty: By default, shell pipelines return the exit status of the last command in the pipeline. Can we do the same here?

src/helpers.c
```diff
@@ -181,9 +181,93 @@ int system_argv(char *const argv[])
...
+{
+	int child_exit_status;
+	pid_t pid;
+	int status = -1;
 
-		return status;
+	pid = fork();
+
+	if (pid == 0) { /* child */
+		if(newstdin >= 0) {
```
phmccarty: There should be a space after the "if" here, and on similar lines below.

phmccarty:
Those are all the comments I have for now on this patch :-)

src/manifest.c
phmccarty, line 866: Never mind this comment... I was looking at the "ret = 0" on the line below, but this patch removes that line, so this is a non-issue.

phmccarty, line 866: The only thing is that "exit" is an unused label now, so that can be removed.

src/chroot.c
phmccarty: I'm still seeing this code change here that won't apply against clearlinux/swupd-server master. Can you try the rebase again?

phmccarty: Ignore these comments. I commented in the wrong place again :-P

phmccarty:
+1, will merge shortly.

phmccarty:
I squashed the series and pushed to master.

PR #17, opened 2016-05-03
Fix bundle manifest hashes from MoM

tmarcu:
The bundle manifest hashes were calculated from the STATE_DIR Manifest.bundle
files, which caused a problem when they the permissions/owner were changed. The
hashes should be calculated by untarring the Manifest.BUNDLE.tar files
referenced in the Manifest.MoM.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -729,6 +729,9 @@ static int write_manifest_plain(struct manifest *manifest)
 	char *conf = config_output_dir();
 	char *filename = NULL;
 	char *submanifest_filename = NULL;
+	char *manifest_tempdir = NULL;
+	char *tarcommand = NULL;
+	char *tempmanifest = NULL;
 	int ret = -1;
 
 	if (conf == NULL) {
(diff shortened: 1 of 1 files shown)


src/manifest.c
```diff
@@ -776,24 +779,49 @@ static int write_manifest_plain(struct manifest *manifest)
...
 
 		string_or_die(&submanifest_filename, "%s/%i/Manifest.%s", conf, file->last_change, file->filename);
 		populate_file_struct(file, submanifest_filename);
-		ret = compute_hash(file, submanifest_filename);
+
+		/* Untar the Manifest.BUNDLE.tar and calculate the hash on that,
+		 * otherwise we may get incorrect hashes due to owner permissions on
+		 * the files. */
+		string_or_die(&tarcommand, TAR_COMMAND " -C %s " TAR_PERM_ATTR_ARGS " -xf %s.tar 2> /dev/null",
+		manifest_tempdir, submanifest_filename);
```
bryteise: Is that the indentation you get from clang-format? Seems off.

bryteise:
+1

src/manifest.c
tmarcu: Hmm looked ok before I sent it.. I see what you mean though. Will recreate the pull request if needed once Patrick checks it out.

src/manifest.c
```diff
@@ -776,24 +779,49 @@ static int write_manifest_plain(struct manifest *manifest)
 	}
 
 	list = g_list_first(manifest->manifests);
+	string_or_die(&manifest_tempdir, "./XXXXXX");
```
phmccarty: Can this tempdir be created in the state directory instead? So, by default, the template would be /var/lib/update/XXXXXX

src/manifest.c
phmccarty: Yes, the indentation looks off for the second line.. manifest_tempdir on the second line should be aligned with &tarcommand on the first.

src/manifest.c
```diff
@@ -776,24 +779,49 @@ static int write_manifest_plain(struct manifest *manifest)
...
+		free(tempmanifest);
 
 		fprintf(out, "%s\t%s\t%i\t%s\n", file_type_to_string(file), file->hash, file->last_change, file->filename);
 		free(submanifest_filename);
 	}
 
 	ret = 0;
 exit:
+	if (rmdir(manifest_tempdir) != 0) {
+		LOG(NULL, "rmdir failed for %s: %d", manifest_tempdir, errno);
```
phmccarty: The string returned by strerror(errno) will be more helpful than errno itself.

src/manifest.c
```diff
@@ -776,24 +779,49 @@ static int write_manifest_plain(struct manifest *manifest)
 	}
 
 	list = g_list_first(manifest->manifests);
+	string_or_die(&manifest_tempdir, "./XXXXXX");
+	if (!mkdtemp(manifest_tempdir)) {
+		LOG(NULL, "Failed to create temporary directory %s for untarring Manifests", manifest_tempdir);
```
phmccarty, line 784: logging the value of strerror(errno) will be useful here too

src/manifest.c
```diff
@@ -776,24 +779,49 @@ static int write_manifest_plain(struct manifest *manifest)
...
+		assert(0);
+	}
+
 	while (list) {
 		file = list->data;
 		list = g_list_next(list);
 
 		string_or_die(&submanifest_filename, "%s/%i/Manifest.%s", conf, file->last_change, file->filename);
 		populate_file_struct(file, submanifest_filename);
```
phmccarty: The populate_file_struct() should be called for tempmanifest instead. In which case, you can simplify and run string_or_die() once and not need the submanifest_filename string at all.

src/manifest.c
tmarcu: You mean using the pre-defined STATE_DIR as the template? 'Cause it won't always be /var/lib/update. However, I don't see the point of it being placed in any specific directory, as all of the contents & tempdir are already wiped before update creation ends, so there's no artifacts left over from it.

phmccarty: By state directory, I mean the configurable state directory (defaults to /var/lib/update). I simply think it's better for swupd-server to always be writing in the state directory whenever possible. One analogous example for using temporary directories is pack content being staged in STATE_DIR/packstage, and then removed after the packs are created. You're right that swupd-server _could_ write in the current directory, or in /tmp, etc. but the state directory is IMO the best place the keep temporary state.

tmarcu:
Updated branch with appropriate changes.

phmccarty:
This change looks good. I followed up with a fix for a couple functional tests that add deleted entries to the MoM, to make sure we don't try to untar nonexistent bundle manifests.

PR #18, opened 2016-05-04
Removing illegal character list for filenames.

jrguzman-intel:
After replacing system() for system_argv() calls, no more filter for special characteres needed for filenames.

->
This patch must be merged after pull request #16 
<-

### DIFF
--- a/src/analyze_fs.c
+++ b/src/analyze_fs.c
@@ -272,35 +272,6 @@ static void get_hash(gpointer data, gpointer user_data)
 	free(filename);
 }
 
-/* disallow characters which can do unexpected things when the filename is
- * used on a tar command line via system("tar [args] filename [more args]");
- */
-static bool illegal_characters(char *filename)
-{
-	char c;
-	int i;
-#define BAD_CHAR_COUNT 11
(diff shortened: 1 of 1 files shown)


phmccarty:
+1, pending the changes from #16 being merged.

phmccarty:
This change also requires a format bump, because the system() calls need removing in swupd-client first, as addressed in clearlinux/swupd-client#74

phmccarty:
The blocking client changes are now tracked at clearlinux/swupd-client#144.

phmccarty:
I'm reconsidering this patch. It's true that by removing the use of `system()` in swupd-client, the special characters interpreted by the shell no longer need to be filtered out from manifests. But I think we should still ban characters if they compromise manifest structure.

Namely, I think we should ban `'\n'`, since the files in a manifest are newline-separated.

tmarcu:
https://github.com/clearlinux/swupd-server/pull/16 must be discussed first.

tmarcu:
Not merging due to https://github.com/clearlinux/swupd-client/pull/144.

phmccarty:
@tmarcu I think we need to open a new issue, because the illegal character list requires revision. At minimum, we need to add `'\n'` to the list, because its presence would corrupt manifest structure.

PR #20, opened 2016-05-10
Remove rsync from full->bundle chroots

tmarcu:
This is no longer needed with the current setup, and is just wasting time during
update creation.


phmccarty:
Looks good, merged.

PR #21, opened 2016-05-16
Parallelize manifest creation and remove deltas

tmarcu:
The code previously created all the bundle manifests and delta manifests
synchronously in 1 thread. While this worked, it was not optimal considering
the manifests can be created at the same time since they are unique. This
patch creates a threadpool that processes each bundle manifest in its own
thread. Delta manifest creation was removed from create_update because it
can already be handled by our delta pack creation process, and it is a huge
time-sink for create_update as it also runs single threaded.

### DIFF
--- a/src/analyze_fs.c
+++ b/src/analyze_fs.c
@@ -481,7 +481,8 @@ struct manifest *sub_manifest_from_directory(char *component, int version)
 	return manifest;
 }
 
-/* get hashes out of full manifest and add them into the component manifest */
+/* Get hashes out of full manifest and add them into the component manifest.
+ * Full manifest must be sorted before calling this function to perform properly */
 void add_component_hashes_to_manifest(struct manifest *compm, struct manifest *fullm)
 {
 	GList *list1, *list2;
@@ -492,7 +493,6 @@ void add_component_hashes_to_manifest(struct manifest *compm, struct manifest *f
 	assert(fullm);
 
 	compm->files = g_list_sort(compm->files, file_sort_filename);
-	fullm->files = g_list_sort(fullm->files, file_sort_filename);
 
 	list1 = g_list_first(compm->files);
 	list2 = g_list_first(fullm->files);
--- a/src/create_update.c
+++ b/src/create_update.c
@@ -39,6 +39,11 @@
 
 #include "swupd.h"
 
+#define NUM_THREADS (sysconf(_SC_NPROCESSORS_ONLN) * 3)
+
+G_LOCK_DEFINE_STATIC(MOM);
+G_LOCK_DEFINE_STATIC(SUBTRACT);
+
 static void banner(void)
 {
 	printf(PACKAGE_NAME " update creator version " PACKAGE_VERSION "\n");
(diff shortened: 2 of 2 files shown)


src/analyze_fs.c
```diff
@@ -492,7 +493,9 @@ void add_component_hashes_to_manifest(struct manifest *compm, struct manifest *f
 	assert(fullm);
 
 	compm->files = g_list_sort(compm->files, file_sort_filename);
-	fullm->files = g_list_sort(fullm->files, file_sort_filename);
+	/* This is a destructive operation so it cannot be threaded easily,
+	 * but it is not needed at this point */
+//	fullm->files = g_list_sort(fullm->files, file_sort_filename);
```
phmccarty: Is it worth keeping this commented code, or should we remove it entirely?

src/create_update.c
```diff
@@ -401,73 +530,15 @@ int main(int argc, char **argv)
...
-		}
-
-		nest_manifest(new_MoM, newm);
-	}
+	manifests->new_full = new_full;
+	manifests->old_core = old_core;
+	manifests->new_core = new_core;
+	manifests->old_manifests = old_manifests;
+	manifests->new_manifests = new_manifests;
+	manifests->manifests_last_versions_list = manifests_last_versions_list;
```
phmccarty, line 539: Looks like os-core manifest deltas are still being created. Can that be removed too?

phmccarty, line 539: on line 489

src/create_update.c
```diff
@@ -212,6 +225,120 @@ static int check_build_env(void)
...
+
+
+	printf("Manifest threadpool %ld threads\n", sysconf(_SC_NPROCESSORS_ONLN) * 3);
+	threadpool = g_thread_pool_new(create_manifest_task, manifests,
+					sysconf(_SC_NPROCESSORS_ONLN) * 3,
+					TRUE, NULL);
+
+	printf("Starting delta manifest creation\n");
+	group = next_group();
+	while(group != NULL) {
```
phmccarty: space after while

src/create_update.c
```diff
@@ -212,6 +225,120 @@ static int check_build_env(void)
...
+	int ret = 0;
+	int count = 0;
+	GError *err = NULL;
+	char *group;
+
+
+	printf("Manifest threadpool %ld threads\n", sysconf(_SC_NPROCESSORS_ONLN) * 3);
+	threadpool = g_thread_pool_new(create_manifest_task, manifests,
+					sysconf(_SC_NPROCESSORS_ONLN) * 3,
+					TRUE, NULL);
```
phmccarty: Is it worth assigning the sysconf() result to a variable, in case we want to change it?

src/create_update.c
```diff
@@ -212,6 +225,120 @@ static int check_build_env(void)
...
+
+	printf("Starting delta manifest creation\n");
```
phmccarty: this message needs updating

src/create_update.c
```diff
@@ -212,6 +225,120 @@ static int check_build_env(void)
...
+	GThreadPool *threadpool;
+	int ret = 0;
+	int count = 0;
+	GError *err = NULL;
+	char *group;
+
+
+	printf("Manifest threadpool %ld threads\n", sysconf(_SC_NPROCESSORS_ONLN) * 3);
+	threadpool = g_thread_pool_new(create_manifest_task, manifests,
+					sysconf(_SC_NPROCESSORS_ONLN) * 3,
```
phmccarty: minor issue, but the type of the third argument is "gint", but sysconf() returns a long int

src/analyze_fs.c
tmarcu: yeah we can remove this, the comment for the function def should be enough.

src/create_update.c
tmarcu, line 539: Removing that as it will be handled by the pack_maker.

phmccarty:
Looks good, will merge shortly.

phmccarty:
I messed up the merge, so I ended up pushing the original revision of this patch, and added a followup patch with the diff to current.

PR #24, opened 2016-05-27
Properly initialise includes for new manifests

incandescant:
Initialise the includes member to null during alloc_manifest,
otherwise when trying to access this member in a setup where
manifest includes aren't used we segfault.

Signed-off-by: Joshua Lock joshua.g.lock@intel.com

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -100,6 +100,7 @@ struct manifest *alloc_manifest(int version, char *component)
 	manifest->version = version;
 	manifest->component = strdup(component);
 	manifest->format = format;
+	manifest->includes = NULL;
 
 	return manifest;
 }


incandescant:
Apologies, I pulled the trigger on this too soon. This is a hunch I'm currently testing when trying to root cause the issue with segfaults mentioned in #22 (it sure smells like an uninitialised variable).

I'll comment here, or close the PR, once my testing is done.

ikeydoherty:
Walk me through this please, that's a `calloc` call, the memory is already zero'd. There's no need that I can determine for the NULL assignment.

The real place the NULL assignment is missing is here:
https://github.com/clearlinux/swupd-server/blob/master/src/manifest.c#L93

```
struct manifest *manifest;
```

should become:

```
struct manifest *manifest = NULL;
```

ikeydoherty:
Do we not have a gdb trace for #22 @incandescant ?

incandescant:
Thanks for the swift feedback @ikeydoherty, I should definitely have been paying more attention.

I hadn't meant to hit the "Create pull request" button before I'd finished testing... 

Essentially this was a stab in the dark attempt to fix an issue we're seeing where swupd_create_update segfaults in our Yocto Project integration, we aren't using manifest includes (yet) and @igor-stoppa traced the segfault as discussed in #22.

I'm still looking at this so will close the PR for now.

ikeydoherty:
@incandescant all good - I noticed the PR buttons/UX got a change overnight anyway.

PR #25, opened 2016-06-10
Resort to lstat for FS not supporting dirent.d_type

rojkov:
According to POSIX.1 only d_name and d_ino fields of struct
dirent are standardized. d_type isn't always correctly set on file
systems like XFS. In such cases it makes sense to resort to
lstat(). Otherwise a user has hard time figuring out what's
wrong with her setup.

### DIFF
--- a/src/analyze_fs.c
+++ b/src/analyze_fs.c
@@ -321,7 +321,6 @@ static void iterate_directory(struct manifest *manifest, char *pathprefix,
 	while (dir) {
 		struct file *file;
 		char *sub_filename;
-		char *fullname;
 
 		entry = readdir(dir);
 		if (!entry) {
@@ -349,12 +348,22 @@ static void iterate_directory(struct manifest *manifest, char *pathprefix,
 		file->last_change = manifest->version;
 		file->filename = sub_filename;
 
-		string_or_die(&fullname, "%s/%s", fullpath, entry->d_name);
-		populate_file_struct(file, fullname);
-		free(fullname);
-
 		if (entry->d_type == DT_DIR) {
 			iterate_directory(manifest, pathprefix, file->filename, do_hash);
+		} else if (entry->d_type == DT_UNKNOWN) { /* fall back to lstat() */
+			struct stat sb;
+			char *fullname;
+
+			string_or_die(&fullname, "%s/%s", fullpath, entry->d_name);
+			if (lstat(fullname, &sb) == -1) {
+				perror("lstat");
+				assert(0);
+			}
+			free(fullname);
+
+			if (S_ISDIR(sb.st_mode)) {
+				iterate_directory(manifest, pathprefix, file->filename, do_hash);
+			}
 		}
 
 		/* if for some reason there is a file in the official build


src/analyze_fs.c
```diff
@@ -351,12 +351,23 @@ static void iterate_directory(struct manifest *manifest, char *pathprefix,
 
 		string_or_die(&fullname, "%s/%s", fullpath, entry->d_name);
 		populate_file_struct(file, fullname);
-		free(fullname);
 
 		if (entry->d_type == DT_DIR) {
 			iterate_directory(manifest, pathprefix, file->filename, do_hash);
+		} else if (entry->d_type == DT_UNKNOWN) { /* fall back to lstat() */
+			struct stat sb;
```
bradTpeters, line 354: populate_file_struct() calls lstat(); why do we need to call it again here?

rojkov, line 354: I didn't notice this `populate_file_struct()`. Why it's needed there then? I presumed it's called for every file in a separate thread (see `get_hash()`) to make these computations parallel. Otherwise there's no need for the `DT_DIR` optimization in the first place.

bradTpeters, line 354: Well it's populating the 'file' struct with the lstat details. That being the case, wouldn't it be easier to just  call

if (S_ISDIR(file.st_mode)) { 

and remove the new call to lstat?

rojkov, line 354: By referring to `get_hash()` I wanted to say that in the current code `populate_file_struct()` is called twice for every file: the first time it's called sequentially and then it's called in the parallel threads again. Which is probably not what you want. AFAIU this

```
if (entry->d_type == DT_DIR) {
```

check is an optimization postponing `lstat()` calls to the parallel threads.

So there are two solutions:
1. use `if (S_ISDIR(file.st_mode)) {`, remove the new call to `lstat()` and remove `populate_file_struct()` from `get_hash()`. This is easier to understand, but suboptimal since `lstat()` is called sequentially for all files.
2. drop the first call to `populate_file_struct()`. In this case the second `lstat()` is called only as a fallback for rare cases where `DT_DIR` is not supported.

bradTpeters:
I think option 2 is indeed the best approach; please revise your patch and resubmit and I'll +1

rojkov:
- removed redundant call to `populate_file_struct()`;
- rebased and force pushed.

phmccarty:
Looks good to me. +1

phmccarty:
For the record:

In master, this change was reverted due to test suite failures. It caused file types to be calculated for Manifest.full, but not for bundle manifests.

phmccarty:
The performance improvements from the earlier discussion on this PR are still worth considering.

In the interim, @tmarcu pushed 74c8f86 to address the d_type portability issue.

pohly:
@phmccarty with performance improvements you mean removing the extra populate_file_struct()? Should this be tracked in a new issue?

phmccarty:
@pohly Exactly. I opened #39 for tracking.

PR #26, opened 2016-06-21
Ensure os-core bundle is listed in groups.ini

incandescant:
As discussed in pull #22 this patch adds a simple check to ensure that the groups.ini file contains the os-core bundle and exits, informing the user the bundle is required, if it does not.

### DIFF
--- a/src/create_update.c
+++ b/src/create_update.c
@@ -212,6 +212,25 @@ static int check_build_env(void)
 	return 0;
 }
 
+static int check_group_file(void)
+{
+	int ret = -1;
+
+	// Ensure the os-core group is defined
+	while (1) {
+		char *group = next_group();
+		if (!group) {
+			break;
+		}
+		if (strcmp(group, "os-core") == 0) {
+			ret = 0;
+			break;
+		}
+	}
+
+	return ret;
+}
+
 int main(int argc, char **argv)
 {
 	struct manifest *new_core = NULL;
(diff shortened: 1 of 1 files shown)


src/create_update.c
```diff
@@ -227,6 +227,25 @@ static int check_build_env(void)
 	return 0;
 }
 
+static int check_group_file(void)
+{
+	int ret = -1;
+
+	// Ensure the os-core group is defined
+	while (1) {
```
bradTpeters, line 220: Just from a stylistic perspective, a while (ret) would be preferable to a while (1) with a break;

bradTpeters:
Other than the conflicts which must be resolved and my style input, looks good

src/create_update.c
incandescant, line 220: Sure, will do. while (1) is pretty common throughout the code base and I was trying to match implicit style but I'll fix this and rebase.

Thanks for the feedback!

incandescant:
Updated to rebase on master and resolve conflicts.

I've left the `while (1)` for now because checking the return value of `next_group()` still requires we break out of the loop early somehow if it returns NULL, requiring one of either a break or early return.

Perhaps this is why other call sites of `next_group()` use a `while (1)` rather than a `while (ret)`?

If `while (ret)` is still the preferred style I can follow on with a change like:

```
    while (ret) {
        char *group = next_group();
        if (!group) {
            break;
        }
        if (strcmp(group, "os-core") == 0) {
            ret = 0;
        }
    }
```

phmccarty:
This change looks good to me. +1

PR #27, opened 2016-07-19
Update manifest creation

tmarcu:
This patch set reverts the parallelization of manifest creation, but keeps the relevant changes removing delta manifest creation.


phmccarty:
+1

PR #30, opened 2016-08-11
Enable the rename feature

phmccarty:
This patch series enables the "rename" feature in swupd-server and can used by passing the --enable-rename-detection option to configure.


phmccarty:
Rebased to latest master, and added an additional change to configure.ac/Makefile.am to ensure that the functional test suite passes regardless of whether `--enable-rename-detection` is used or not.

tmarcu:
New PR available.

PR #31, opened 2016-09-07
Document and ship bundle status string.

sofar:
This patch extracts the [STATUS] field from our bundle information
metadata, and stores it in groups.ini. From there we put the contents
of this string verbatim into the manifest.

We don't interpret, encode or convert the contents of the [STATUS]
field in the manifest. Instead, we just strip non-alphanumeric
characters and pass the contents on. This leaves it entirely to the
client to parse and interpret the value of this field in the manifest.

If the bundle file, or the groups.ini file omits any status, nothing
is output to the Manifest file.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -37,6 +37,7 @@ swupd_make_pack_SOURCES = \
 	src/config.c \
 	src/delta.c \
 	src/globals.c \
+	src/groups.c \
 	src/helpers.c \
 	src/log.c \
 	src/make_packs.c \
@@ -53,6 +54,7 @@ swupd_make_fullfiles_SOURCES = \
 	src/delta.c \
 	src/fullfiles.c \
 	src/globals.c \
+	src/groups.c \
 	src/helpers.c \
 	src/log.c \
 	src/make_fullfiles.c \
--- a/include/swupd.h
+++ b/include/swupd.h
@@ -221,6 +221,7 @@ extern void read_group_file(char *filename);
 extern void release_group_file(void);
 extern char *group_groups(char *group);
 extern char *group_packages(char *group);
+extern char *group_status(char *group);
 extern char *next_group(void);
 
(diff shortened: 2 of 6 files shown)


sofar:
Testing: This was tested using the `make check` code which validates that output manifests correctly read `groups.ini`. The `make check` code omits testing the conversion from bundle files to `groups.ini`, but that part is trivial enough and was tested on the command line, since it's a one-liner.

Makefile.am
```diff
@@ -37,6 +37,7 @@ swupd_make_pack_SOURCES = \
 	src/config.c \
 	src/delta.c \
 	src/globals.c \
+	src/groups.c \
```
phmccarty, line 40: I don't think swupd_make_pack needs this source file, since swupd_create_update is the only caller of `write_manifest_plain()`.

Makefile.am
```diff
@@ -53,6 +54,7 @@ swupd_make_fullfiles_SOURCES = \
 	src/delta.c \
 	src/fullfiles.c \
 	src/globals.c \
+	src/groups.c \
```
phmccarty, line 57: Same comment as above, though for swupd_make_fullfiles.

phmccarty:
The code changes look okay to me. I'll check swupd-client to see if it skips unknown fields in manifest headers.

phmccarty:
swupd-client does skip header entries it doesn't know about...

+1

Makefile.am
sofar: Patrick and me dug into this, but with groups.c symbols used in manifest.c, we can't avoid this unless we shuffle a bunch of code around first. So this should be OK.

phmccarty:
After further examination, I see manifest.c is needed by the other binaries, so groups.c must be included. So disregard my two inline comments.

tmarcu:
Merged 7758a6

PR #34, opened 2016-10-12
RFC: Ostro enhancements

pohly:
This is the result of my work on addressing the performance problems that Ostro had with swupd. Where previously image creation took over 4 hours, it's now down to a bit more than half an hour. Not all of that came from these swupd-server enhancements, but a significant part.

The first four patches are hopefully uncontroversial and ready to be merged; I can submit them separately in a PR if you agree, or you can just merge manually from my branch.

The commit that will require more thought and/or work is "fullfiles: use libarchive directly". Right now this change is unconditional, i.e. it also changes the output when bsdtar is not explicitly requested during compilation. I did that because the new code is simpler and you might want to use it instead of the old one. My hope is that this won't cause problems in practice for Clear because GNU tar on Clear might (should?) be able to unpack the .tar archives without problems. That's based on the assumption that you don't have xattrs (which are known to be stored in an incompatible way) and that long file names are okay. But I haven't tested that. Can you check that using some real Clear build?

If you can't and/or don't want to change, would a patch with ifdefs be acceptable, i.e. only call libarchive directly when bsdtar is enabled?

"swupd-create-update: alternative input layout" should not change anything when using the traditional input layout. The patch itself is a bit larger because it moves some code into the add_file() utility method.

Finally, the manifest hash patches ("swupd_create_update: call external command for each manifest" and "swupd-server: include xattrs in manifest hash"): they are necessary if we want to check xattrs set for manifest files. I'm undecided myself whether that's really needed. It might be simpler to just patch the swupd-client's verify_file() method.

However, setting the security.IMA xattr via an external command might have some advantages even when not checking the xattr, so I'd prefer to get at least that one merged.

@rojkov @incandescant @igor-stoppa: FYI, this patch review is required before https://github.com/ostroproject/ostro-os/pull/198 can proceed.

### DIFF
--- a/include/swupd.h
+++ b/include/swupd.h
@@ -20,12 +20,12 @@
 #define TAR_COMMAND "bsdtar"
 #define TAR_XATTR_ARGS ""
 #define TAR_XATTR_ARGS_STRLIST
-#define TAR_WARN_ARGS ""
+#define TAR_WARN_ARGS_STRLIST
 #else
 #define TAR_COMMAND "tar"
 #define TAR_XATTR_ARGS "--xattrs --xattrs-include='*'"
 #define TAR_XATTR_ARGS_STRLIST "--xattrs", "--xattrs-include='*'",
-#define TAR_WARN_ARGS "--warning=no-timestamp"
+#define TAR_WARN_ARGS_STRLIST "--warning=no-timestamp",
 #endif
 
 #if SWUPD_WITH_SELINUX
--- a/include/swupd.h
+++ b/include/swupd.h
@@ -147,12 +147,14 @@ extern char *state_dir;
 extern char *packstage_dir;
 extern char *image_dir;
 extern char *staging_dir;
+extern char *manifest_cmd;
 
 extern bool init_globals(void);
 extern void free_globals(void);
 extern bool set_format(char *);
 extern void check_root(void);
 extern bool set_state_dir(char *);
+extern bool set_manifest_cmd(const char *cmd);
 extern bool init_state_globals(void);
 extern void free_state_globals(void);
 
--- a/src/analyze_fs.c
+++ b/src/analyze_fs.c
@@ -387,6 +387,9 @@ struct manifest *full_manifest_from_directory(int version)
 {
 	struct manifest *manifest;
 	char *dir;
+	int numthreads = getenv("SWUPD_NUM_THREADS") ?
+		atoi(getenv("SWUPD_NUM_THREADS")) :
+		sysconf(_SC_NPROCESSORS_ONLN);
 
 	LOG(NULL, "Computing hashes", "for %i/full", version);
 
@@ -394,7 +397,7 @@ struct manifest *full_manifest_from_directory(int version)
 
 	string_or_die(&dir, "%s/%i/full", image_dir, version);
 
-	threadpool = g_thread_pool_new(get_hash, dir, 12, FALSE, NULL);
+	threadpool = g_thread_pool_new(get_hash, dir, numthreads, FALSE, NULL);
 
 	iterate_directory(manifest, dir, "", true);
 
--- a/src/analyze_fs.c
+++ b/src/analyze_fs.c
@@ -275,7 +275,7 @@ static void get_hash(gpointer data, gpointer user_data)
 /* disallow characters which can do unexpected things when the filename is
  * used on a tar command line via system("tar [args] filename [more args]");
  */
-static bool illegal_characters(char *filename)
+static bool illegal_characters(const char *filename)
 {
 	char c;
 	int i;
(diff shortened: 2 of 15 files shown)


pohly:
@phmccarty ping?

bryteise:
@tmarcu is the new maintainer so just want to make sure he gets notified as well. I will really try and get some time to review this too.

bryteise:
As an aside the commit messages add some extra metadata that isn't particularly helpful for the swupd repo to contain. Would you mind that being trimmed off?

pohly:
@bryteise yes, I'll remove these extra comments. They came from the tooling that was used to develop the patch series and I'm so used to them that I mentally blanked them out when reviewing my patches before submission ;-}

src/analyze_fs.c
```diff
@@ -387,14 +387,17 @@ struct manifest *full_manifest_from_directory(int version)
 {
 	struct manifest *manifest;
 	char *dir;
+	int numthreads = getenv("SWUPD_NUM_THREADS") ?
+		atoi(getenv("SWUPD_NUM_THREADS")) :
+		sysconf(_SC_NPROCESSORS_ONLN);
```
bryteise, line 475: Since the numthreads value is passed right into g_thread_pool_new it might be best to handle invalid values (-1 is understood to mean unlimited in this case so seems valid though in fullfiles, -1 and exclusive being set to TRUE would cause it to fail as well) as we don't handle errors from the threadpool otherwise from the look of it.

bryteise: Also since this code is duplicated in three places it would seem reasonable to throw it in a macro/function I think.

include/swupd.h
```diff
@@ -20,12 +20,12 @@
...
 #define TAR_XATTR_ARGS ""
 #define TAR_XATTR_ARGS_STRLIST
-#define TAR_WARN_ARGS ""
+#define TAR_WARN_ARGS_STRLIST
 #else
 #define TAR_COMMAND "tar"
 #define TAR_XATTR_ARGS "--xattrs --xattrs-include='*'"
 #define TAR_XATTR_ARGS_STRLIST "--xattrs", "--xattrs-include='*'",
-#define TAR_WARN_ARGS "--warning=no-timestamp"
+#define TAR_WARN_ARGS_STRLIST "--warning=no-timestamp",
```
bryteise, line 28: Okay so this confuses me sorry. From the commit message I was thinking that bsdtar required the comma not tar (since that is what is being used today without modification) but this is a change to the tar invocation too.

So the patch is fixing both the bsdtar case and tar case which are both broken (but doesn't seem to be an issue for how Clear Linux is currently being built)?

src/log.c
```diff
@@ -33,21 +33,21 @@
...
-	logfile = fopen(filename, "w");
+	logfile[0] = fopen(filename, "w");
 	free(filename);
 	gettimeofday(&start_time, NULL);
 }
 void init_log_stdout(void)
 {
-	logfile = stdout;
+	logfile[1] = stdout;
 	gettimeofday(&start_time, NULL);
```
bryteise, line 51: It'd be nice to not init start_time twice but it likely isn't too important.

bryteise:
The libarchive patches (4ebc886 and 9d85a80) are likely going to be controversial based on confidence in libarchive performing correctly being over all low. I'm holding off on reviewing those for now.

src/analyze_fs.c
```diff
@@ -275,7 +275,7 @@ static void get_hash(gpointer data, gpointer user_data)
 /* disallow characters which can do unexpected things when the filename is
  * used on a tar command line via system("tar [args] filename [more args]");
  */
-static bool illegal_characters(char *filename)
+static bool illegal_characters(const char *filename)
```
bryteise, line 278: Unrelated but good change.

bryteise:
The alternative input patch (37b5f5e) looks interesting. Very curious about it from a design standpoint but I'm going to spend more time reviewing it so putting it off for now.

bryteise:
For the call command for each manifest patch (38a3a23), is there a reason that this can't be part of the deployment process outside of swupd instead? It seems to be very one off and more of a per use implementation process.

src/manifest.c
```diff
@@ -812,6 +812,11 @@ static int write_manifest_plain(struct manifest *manifest)
 		}
 
 		string_or_die(&tempmanifest, "%s/Manifest.%s", manifest_tempdir, file->filename);
+		/*
+		 * use_xattr Has to match swupd-client:
+		 * there it is enabled unconditionally in verify_file().
+		 */
+		file->use_xattrs = true;
```
bryteise, line 819: Oh that's interesting it wasn't matching before. This seems correct to me regardless of anything else then.

bryteise:
Okay, my review is complete minus the noted libarchive and full directory patches (the latter I'll get to hopefully sometime soon).

src/analyze_fs.c
pohly: I agree, a function would be better. But how would error handling/reporting work? Just exit()?

Is there enough interest in this enhancement to spend more time on this? I only used it while debugging, and now could drop the entire patch again, too.

include/swupd.h
pohly: The old usage was:

```
char *const tarcmd[] = { TAR_COMMAND, "-C", param, TAR_WARN_ARGS, TAR_PERM_ATTR_ARGS_STRLIST, "-xf", path, NULL };
```

That worked for GNU tar because TAR_WARN_ARGS was exactly one parameter (i.e. neither empty nor more than one).

It did not work for bsdtar because TAR_WARN_ARGS was empty, inserting an additional, empty parameter into the command's argv. For defines that may contain a varying number or parameters with have to use the _STRLIST approach with explicit commas.

pohly:
Low confidence in libarchive performing correctly? Have you had negative experiences with it, or is it just the lack of actual experience that makes you wary about switching to it?

FWIW, my own experience with it was rather positive. bsdtar's behavior when restoring files is much saner than GNU tar's (less syscalls), the code is well structured, adding a new feature (allowing -n in combination with -x and -t) was straight-forward, upstream is responsive, good support for the more modern PAX format (GNU tar wants to switch to that, but hasn't yet).

GNU tar on the other hand surprised me recently with a apparently random bug in the -T implementation. I haven't looked at the source code, so I don't know how good the code is.

src/analyze_fs.c
pohly: The change was necessary because the new call to the function passes a const string.

src/manifest.c
pohly: Yes, it was basically broken before on systems where xattrs get set at runtime (Smack, IMA, SE Linux (?)).

Regarding the proposed fix of calling an external command on the server side: this can't be handled via some kind of external deployment, because the manifest files get created by swupd-server and then get hashed immediately. The necessary xattrs have to be set between these two steps, hence the external command.

The alternative solution is to patch swupd-client so that it does not check xattrs. I'm open for that, too, because I can't think of any good reason why checking xattrs for these data files provides additional value that would justify the complexity. But perhaps I'm just missing such a reason (additional protection against tampering when signed IMA hashes are used perhaps?).

pohly:
Let's finish the current review round before I rework the patches, which then will also remove the extra lines from the commit messages.

@bryteise thanks for your time. @tmarcu, now would be a good time to chime in.

src/analyze_fs.c
bryteise: So I think we could just have the function return 1 on error. I don't see anything terrible about that since it should just always work if not as fast.

include/swupd.h
bryteise: Ah okay that makes sense and seems like good future proofing.

src/analyze_fs.c
bryteise: Ah I see that the strrchr is casted now, thanks!

src/manifest.c
bryteise: Ah I like the proposal of not doing the check for xattrs on Manifest files in the client. That seems to make more sense to me because xattrs on the Manifest files do not make sense to me.

bryteise:
@pohly My understanding was that the first implementation of swupd tried to use libarchive and had correctness problems relating to it. If @tpepper can recall specific things that would be nice but I'd also be extremely happy if libarchive works for you in practice for some time and swupd can migrate away from running tar commands on client and server.

sofar:
I'm in favor of using `libarchive` myself. The project is actively maintained and very active as well, so if issues should happen, there's at least a place to get them resolved.

src/manifest.c
icarus-sparry: A more general solution is to change the manifest format to list the attributes which should be checked.

pohly: @icarus-sparry this sounds like a rather complex solution (these server-side changes plus additional code for this format extension) for no obvious gain. The only advantage that I could come up with for checking xattrs on Manifest files was that signed security.ima provides some additional integrity protection. But that's redundant, swupd itself already does that.

pohly:
Replaced by individual PRs.

PR #36, opened 2016-11-07
swupd_create_fullfiles: avoid segfault when nothing changes

pohly:
In the (unlikely) case that nothing changed between two builds,
get_deduplicated_fullfile_list() segfaults because it uses
manifest->files without checking for NULL, aka the empty list.


phmccarty:
+1

PR #37, opened 2016-11-08
delta.c: fix xattr test after patching

pohly:
At the moment, swupd_create_pack fails when some files have xattrs and
get patched because the xattrs of the test file do not match the
original, unpatched file.

That's because xattrs_copy() was applied to the wrong target file.

Fixes: swupd-server/#35


phmccarty:
Good catch! +1 from me.

PR #40, opened 2016-11-18
Fix fallthrough state detection logic

phmccarty:
This conditional checks for state *directories* that are generally
installed by default, and the conditional immediately below this one
checks for state files within these directories. So, if we do strncmp()
instead of strcmp(), the fallthrough logic doesn't occur, and state
files are not marked as such.

This reverts commit 63fb5fb61bc95525635deb2b0f4bdbf9d1e0e1ae.

Signed-off-by: Patrick McCarty <patrick.mccarty@intel.com>


phmccarty:
I will add a functional test.

phmccarty:
Added a commit with a functional test.

PR #41, opened 2016-11-18
swupd_make_pack: fix extracting files with bsdtar

pohly:
TAR_XATTR_ARGS is no longer used as part of a plain string. Embedding
the empty "" value for bsdtar inside an argv argument list passes an
empty parameter to bsdtar, leading to:
  bsdtar: Must specify one of -c, -r, -t, -u, -x

To allow the the "no parameter" case, it has to be argument list: that
can be empty. If not empty, it has to end with a comma.

Already reviewed here: https://github.com/clearlinux/swupd-server/pull/34#discussion_r85226517


phmccarty:
+1

PR #42, opened 2016-11-18
fullfiles.c: fix invalid LOG() call

pohly:
LOG() takes an additional fixed string before the format string.

Already reviewed here: https://github.com/clearlinux/swupd-server/pull/34

IMHO it would be good to add printf format function attributes to the
logging functions, to catch such errors and also mismatches between
the format string and the actual parameters.

https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes


phmccarty:
+1

PR #43, opened 2016-11-18
update control over parallelism

pohly:
The SWUPD_NUM_THREADS env variable is now understood by all three
commands and overrides the default number of threads. Setting it to 1
is useful while debugging the code that runs inside threads (only one
thread hits breakpoints there). If SWUPD_NUM_THREADS is invalid, a
warning is printed and the variable gets ignored, i.e. the default
parallelism is used.

The hard-coded parallelism of 12 threads when analysing the file system
gets replaced with n, where n is the number of available CPUs. The default
is the same as before elsewhere (n for packing, 3 * n for fullfiles).

This is includes the proposed change from
https://github.com/clearlinux/swupd-server/pull/34#discussion_r84758929
(separate helper function). Because it was less code this way, the
default number of threads (instead of 1, as suggested) is use in case
of an invalid env var.

### DIFF
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -284,3 +284,28 @@ void check_root(void)
 		exit(EXIT_FAILURE);
 	}
 }
+
+int num_threads(float scaling)
+{
+	const char *var = getenv("SWUPD_NUM_THREADS");
+	int result = sysconf(_SC_NPROCESSORS_ONLN) * scaling;
+
+	if (var && *var) {
+		char *endptr;
+		long int value;
+
+		errno = 0;
+		value = strtol(var, &endptr, 0);
+
+		if ((errno != 0 && value == 0) || *endptr) {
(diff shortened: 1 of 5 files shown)


src/helpers.c
```diff
@@ -284,3 +284,29 @@ void check_root(void)
 		exit(EXIT_FAILURE);
 	}
 }
+
+int num_threads(float scaling)
+{
+	const char *var = getenv("SWUPD_NUM_THREADS");
+	int result = sysconf(_SC_NPROCESSORS_ONLN) * scaling;
+	float scaled_result;
```
phmccarty: This variable appears unused.

pohly: Indeed. Removed.

phmccarty:
+1

tmarcu:
Looks good here.

PR #44, opened 2016-11-18
enable locales in all programs

pohly:
This is a pre-condition for using libarchive directly: libarchive
needs to know what the encoding of filenames is, and it uses the
current locale for that. Without setlocale(), the locale is "C", which
only supports ASCII filenames, leading to warnings about "Can't
encode..." from libarchive when it is forced to fall back to copying
strings verbatim when writing archives that require UTF-8 encoding.

As a side effect, error messages from libc will get translated
according to the user's environment.

I'm submitting this separately from the libarchive changes because I
find it useful also without those and merging it now would reduce the
number of non-upstream patches.

### DIFF
--- a/src/create_update.c
+++ b/src/create_update.c
@@ -29,6 +29,7 @@
 #include <errno.h>
 #include <getopt.h>
 #include <glib.h>
+#include <locale.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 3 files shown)


phmccarty:
Does libarchive need all the locale variables set (LC_ALL), or only certain ones? I'm mostly concerned about LC_COLLATE; to ease testing and reproducibility, I would want manifest sorting order to be predictable.

pohly:
It seems to test LC_CTYPE (get_current_codepage in libarchive/archive_string.c).

Where are manifests getting sorted? Is that code really taking the current locale into account? If that's meant to be predictable, perhaps it should use a string comparison which is locale-independent?

phmccarty:
@pohly The filename sorting uses `strcmp`. I misspoke though, since I thought `strcmp` used LC_COLLATE, but that's `strcoll`.

I'm not seeing any other issues with merging this change.

phmccarty:
+1

PR #45, opened 2016-11-18
swupd-create-update: alternative input layout

pohly:
In Ostro OS, we already have a "full" directory with all files.
Splitting it up into bundles just so that swupd-create-update can
reconstruct the "full" directory is a waste of IO, and noticably slow
when run under pseudo.

To streamline the required work, a new layout for the "image" input
directory gets introduced:
- The "full" directory gets created by the caller before invoking
  swupd-create-update.
- For each bundle, instead of a <bundle> directory, there is a
  <bundle>.content.txt file, listing all entries (including directories)
  of the bundle.

The traditional mode of operation still works as before because each operation
which normally works with a bundle directory checks whether there is such a
directory and if not, switches to the new mode.

That way it is even possible to mix the two modes, i.e. replacing only
some bundles with a content list, although that's probably not all
that useful.

### DIFF
--- a/src/analyze_fs.c
+++ b/src/analyze_fs.c
@@ -275,7 +275,7 @@ static void get_hash(gpointer data, gpointer user_data)
 /* disallow characters which can do unexpected things when the filename is
  * used on a tar command line via system("tar [args] filename [more args]");
  */
-static bool illegal_characters(char *filename)
+static bool illegal_characters(const char *filename)
 {
 	char c;
 	int i;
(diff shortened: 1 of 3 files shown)


src/analyze_fs.c
```diff
@@ -301,27 +301,149 @@ static bool illegal_characters(char *filename)
...
+		 * If there is a <dir>.content.txt instead of
+		 * the actual directory, then read that
+		 * file. It has a list of path names,
+		 * including all directories. The
+		 * corresponding file system entry is then
+		 * expected to be in a pre-populated "full"
+		 * directory.
+		 */
+		if (subpath[0]) {
+			string_or_die(&fullpath, "%s/%s.content.txt", pathprefix, len, subpath);
```
phmccarty: Did you mean for the second `%s` to instead be `%.*s`?

src/analyze_fs.c
```diff
@@ -301,27 +301,149 @@ static bool illegal_characters(char *filename)
...
-	GError *err = NULL;
 
 	string_or_die(&fullpath, "%s/%s", pathprefix, subpath);
 
 	dir = opendir(fullpath);
 	if (!dir) {
+		FILE *content;
+		int len;
 		free(fullpath);
+		if (errno != ENOENT) {
```
phmccarty: Very unlikely, but `free()` may set errno, so errno should be checked before the free.

src/fullfiles.c
```diff
@@ -261,6 +261,8 @@ static GList *get_deduplicated_fullfile_list(struct manifest *manifest)
 			prev = tmp;
 			outfiles = g_list_prepend(outfiles, tmp);
 		}
+		close(fd);
```
phmccarty: I don't see this variable declared.

phmccarty (commented):
Very interesting change! I'll think it through some more, but here are some initial comments.

src/analyze_fs.c
pohly: I'm not sure that free() really sets errno. The man page for free only mentions that allocating may set ENOMEM. Do you have a pointer to documentation which says otherwise?

Having said that, of course it isn't hard to change the code such that allows free to change errno. Will do.

src/analyze_fs.c
pohly: Yes. Clearly doesn't work this way, in particular because len wasn't getting set anyway. I think I had some (now lost) code which analyzes "subpath" when is non-empty.

This is relevant only for the situation where some directory deep inside an otherwise populated directory tree gets replaced with a .content.txt file. That situation isn't needed by meta-swupd, so I also never tested it.

Instead of fixing the unused code, I'll just add an assert() and some explanation that .content.txt is only meant to be used at the top level.

src/fullfiles.c
pohly: That particular "close(fd);" belongs into a different commit. I'll fix that.

pohly:
Updated commit pushed.

tmarcu:
+1 for supporting both modes, I'd like to explore this more for our case as wellin the future, but in the meantime we can merge this.

phmccarty:
Looks fine. +1

tmarcu:
Reverted in master, see issue https://github.com/clearlinux/swupd-server/issues/54.

PR #46, opened 2016-11-18
add logging to stdout

pohly:
When a CI system (like the one from Ostro) captures the output of
commands, but not necessarily intermediate log files, then it is
useful to also log to stdout. Another use case is calling the tools
interactively during development.

The new --log-stdout option in all three commands enables logging to
stdout in addition to the traditional log files.

The implementation recycles the existing init_log_stdout() (not used
before) and gives it the slightly different meaning of "also log to
stdout".

Reviewed before here: https://github.com/clearlinux/swupd-server/pull/34#discussion_r84761155


phmccarty:
+1

Issue #50, opened 2016-11-23
Support type changes from directories to files and symlinks


phmccarty:
Type changes for directory->file and directory->symlink are currently not supported, but it will useful to support them.

phmccarty:
More background posted to the wiki: https://github.com/clearlinux/swupd-server/wiki/Type-changes

pohly:
Can this be extended to also cover file->symlink?

We have that situation in Ostro OS 1.0 -> current release for /usr/lib/ssl/openssl.cnf. I'm not sure what further work is needed besides testing; it seems to work already with swupd client 2.87.

phmccarty:
@pohly file->symlink is already supported. Just the directory->N type changes are unsupported at the moment.

PR #51, opened 2016-11-29
type_change: allow type change from directory to symlink

andred:
At least bsdtar (on the client side) seems to handle
this perfectly fine.

Signed-off-by: André Draszik <git@andred.net>

### DIFF
--- a/src/type_change.c
+++ b/src/type_change.c
@@ -81,11 +81,13 @@ static bool type_has_changed(struct file *file)
 	if (((type1 == 'F') && (type2 == 'L')) ||
 	    ((type1 == 'F') && (type2 == 'D')) ||
 	    ((type1 == 'L') && (type2 == 'F')) ||
-	    ((type1 == 'L') && (type2 == 'D'))) {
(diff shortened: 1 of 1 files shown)


andred:
I hope I didn't miss anything, but this in combination with https://github.com/clearlinux/swupd-client/pull/168 seems to work perfectly fine for me.

andred:
I didn't actually notice https://github.com/clearlinux/swupd-server/issues/50 before, does that issue imply there is more to do?

phmccarty:
@andred Yes, more work is needed to support a non-empty directory type change. The empty directory case is the trivial one.

I just wrote a summary of the current state at https://github.com/clearlinux/swupd-server/wiki/Type-changes.

phmccarty:
Hmm, I see that `swupd_rm()` is called in swupd-client for this case already, so it may work okay after all. I'll revisit the client staging code.

tmarcu:
Refer to https://github.com/clearlinux/swupd-client/issues/209 for further work on this.

Issue #29, opened 2016-08-08
No certificate check in packfsck.pl


johnandersen777:
I'm not clear on what `packfsck.pl` is for, but I don't think it should be using `--no-check-certificate` no matter what it is.

https://github.com/clearlinux/swupd-server/blob/master/packfsck.pl#L10

phmccarty:
The script can be used for validating pack content, since it cross-references the bundle manifests. Currently, it only checks `os-core` but could be made generic to validate any/all packs.

phmccarty:
Fix merged in 490326f.

PR #52, opened 2016-11-30
packfsck: honor proxy settings and cert checking

phmccarty:
In case a proxy is used for downloading the manifest/pack, make sure to
honor those settings. And also, make sure certificate verification
occurs for the downloads.

Signed-off-by: Patrick McCarty <patrick.mccarty@intel.com>


phmccarty:
Addresses #29

PR #57, opened 2017-04-03
Only create deltas between two regular files

phmccarty:
For three different Clear Linux OS builds in the last few months, deltas
were created between files with type change L->F (symlink to file).
This was allowed to occur because there is no check if
file->peer->is_link in __create_delta().

Instead, remove the file->is_link check and simply ensure that the
from/to file types are both F (i.e. "regular file").

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -109,14 +109,14 @@ TESTS = $(dist_check_SCRIPTS)
 dist_check_SCRIPTS = \
 	test/functional/basic/test.bats \
 	test/functional/delete-no-version-bump/test.bats \
-	test/functional/include-version-bump/test.bats \
-	test/functional/update/test.bats \
-	test/functional/fullfiles/test.bats \
-	test/functional/pack/test.bats \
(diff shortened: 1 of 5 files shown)


tmarcu:
+1, test for this would be nice to add. Maybe a TODO would be good so this is updated in the future when we do end up supporting more?

phmccarty:
I will add a test shortly. I don't think it will make sense to create deltas between anything other than regular files...

Note that this code may need some rework for rename support. I didn't look into that.

phmccarty:
I meant to use a logical OR instead of AND... the test I am writing caught this bug. I'll upload a revised patch + test shortly.

phmccarty:
Uploaded a new patch set

tmarcu:
Which parts of this code would have to be taken into consideration for rename support?

phmccarty:
@tmarcu I'm not sure. My comment was more of a heads up that I have no idea how the current rename code creates deltas :-)

PR #58, opened 2017-04-05
Skip subtraction when both files are deleted

phmccarty:
Subtracting files from manifests when both versions are marked deleted proves problematic for client updates, because the version at which the files were deleted will differ, and the client uses the versions to determine when it should delete files for an update.

As an example, consider a distro with 2 bundles, "A" and "B". Bundle B includes bundle A, which results in files in A being subtracted from B.  In this situation the following sequence of four changes result in a subtraction that prevents a client update from deleting a file.

1) In version 10, file /usr/foo is added to bundles A and B.
2) In version 20, /usr/foo is deleted from bundle A.
3) In version 30, /usr/foo is deleted from bundle B.
4) In version 40, bundle B is modified.

Due to arbitrary modifications to bundle B in step 4, /usr/foo is subtracted from bundle B, because it's also deleted in bundle A.  However, the file versions mismatch. So, an update from version 20 to 40 will result in /usr/foo not being deleted because the deleted entry from bundle A was not deleted in a version newer than 20.

This patch fixes the issue by ensuring that the deleted entry in bundle B remains intact for version 30. And the client update then works correctly: an update from 20 to 40 will properly delete /usr/foo, because 20 < 30 <= 40.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -582,6 +582,16 @@ void subtract_manifests(struct manifest *m1, struct manifest *m2)
 			list1 = g_list_next(list1);
 			list2 = g_list_next(list2);
 
+			/* When both files are marked deleted, skip
+			 * subtraction. Preserving the deleted entries in both
+			 * manifests is required for 'swupd update' to know
+			 * when to delete the file, because the m2 bundle may
+			 * be installed with or without the m1 bundle.
+			 */
+			if (file1->is_deleted && file2->is_deleted) {
+				continue;
+			}
+
 			if (file1->is_deleted == file2->is_deleted && file1->is_file == file2->is_file) {
 				m1->files = g_list_delete_link(m1->files, todel);
 				m1->count--;
--- a/Makefile.am
+++ b/Makefile.am
@@ -117,6 +117,7 @@ dist_check_SCRIPTS = \
 	test/functional/no-delta/test.bats \
 	test/functional/pack/test.bats \
 	test/functional/state-file/test.bats \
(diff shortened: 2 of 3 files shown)


src/manifest.c
```diff
@@ -582,6 +582,16 @@ void subtract_manifests(struct manifest *m1, struct manifest *m2)
 			list1 = g_list_next(list1);
 			list2 = g_list_next(list2);
 
+			/* When both files are deleted, skip subtraction, since
```
tmarcu: Can this be re-worded to mention that both files being deleted && having a same last_change value does not qualify them for subtractions still? Also a test for this would be nice :)

tmarcu (changes requested):

src/manifest.c
phmccarty: Sure, I'll update the comment and add a new test.

phmccarty: The same last_change case is more complex to think about. I'll revisit tomorrow.

phmccarty: After thinking about that case some more, I determined that it's currently impossible for the last_change values to be equal in this case. Because both files would be present in the previous version and therefore subtracted, the delete for the current version would be present in the subtracted manifest but not in the manifest subtracted from.

I'll update the code comment for clarification, but no further test case is needed that I can think of right now.

phmccarty:
Uploaded patch set 2, with a revised code comment and slight fixes to the first commit's description.

tmarcu (approved):

tmarcu:
lgtm with the new descriptions.

PR #59, opened 2017-04-10
Avoid duplicate includes lines in manifests

phmccarty:
There is no need for duplicate includes to exist in manifest headers, so
search the includes lists first before adding a new entry.


tmarcu:
+1

PR #64, opened 2017-07-27
Post update action for format bumps

matthewrsj:
When a format bump occurs and the new format is greater than the old
format, an actions field is written to the Manifest.MoM containing the
string "update". This "update" action tells the client that it is
necessary to re-execute swupd update to bring the client to the latest
version within the new format.

Also adds a functional test for the `actions:\tudpate` line in `Manifest.MoM` for a new format.


tmarcu:
+1

PR #65, opened 2017-08-16
Calculate entire contentsize for manifest

matthewrsj:
Instead of only calculating the contentsize for files that were updated
in the current version (an inaccurate number for download size since
this is not the compressed size), calculate for all files in the
manifest. Additionally, do not add submanifest contentsizes to the
current manifest contentsize, as this will result in overcount on client
systems when multiple bundles include the same bundle.

With this change the contentsize field of the manifests will only report
the size of the files unique to that bundle. It is then the client's
responsibility to calculate total bundle size including included
bundles. This is reasonably easy to accomplish with the upcoming
swupd-client bundle-list --deps feature.

Although the above use case is a bit more work for the client, it
additionally allows a user to calculate the installation size of
multiple bundles much more easily, since it only has to count bundle
dependencies once to ensure files are not over counted.

If two bundles not in the same include chain have overlapping
content, summing the include chain of each bundle in the client will result
in an over-estimation of the total size on the system. The more content is
shared, the higher the over-estimation. In reality this overlap will not be
large, but it is currently impossible to calculate the exact installed size
using just the contentsize.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -675,19 +675,29 @@ char *file_type_to_string(struct file *file)
 	return type;
 }
 
+/* Calculate the contentsize for the manifest based on file sizes.
+ *
+ * This should calculate the files uniquely included in this manifest, but none
+ * of its submanifests, which will allow calculation of sizes of all bundles on
+ * a system by adding all manifest->contentsizes of installed bundles.
+ *
+ * However, if two bundles not in the same include chain have overlapping
+ * content, summing the include chain of each bundle in the client will result
+ * in an over-estimation of the total size on the system. The more content is
+ * shared, the higher the over-estimation. In reality this overlap will not be
+ * large, but it is currently impossible to calculate the exact installed size
+ * using just the contentsize.
+ */
 static void compute_content_size(struct manifest *manifest)
 {
-	/* FIXME: this is a temporary implementation based on worst case */
-
 	GList *list;
 	struct file *file;
-	struct manifest *submanifest;
 
 	list = g_list_first(manifest->files);
 	while (list) {
 		file = list->data;
 		list = g_list_next(list);
-		if (!file->is_deleted && (file->last_change == manifest->version)) {
+		if (!file->is_deleted) {
 			if (file->is_file) {
 				manifest->contentsize += file->stat.st_size;
 			} else if (file->is_link) {
@@ -695,17 +705,6 @@ static void compute_content_size(struct manifest *manifest)
 			} else if (file->is_dir) {
 				manifest->contentsize += DIR_SIZE_HINT;
 			}
-		}
-	}
-
-	list = g_list_first(manifest->submanifests);
-	while (list) {
-		submanifest = list->data;
-		list = g_list_next(list);
-
-		/* Do not take into account groups not included in download content */
-		if (create_download_content_for_group(submanifest->component)) {
-			manifest->contentsize += submanifest->contentsize;
 		}
 	}
 }
--- a/test/functional/contentsize-across-versions-includes/test.bats
+++ b/test/functional/contentsize-across-versions-includes/test.bats
@@ -1,0 +1,64 @@
+#!/usr/bin/env bats
+
+# common functions
+load "../swupdlib"
+
+setup() {
+  clean_test_dir
+  init_test_dir
+
+  init_server_ini
+  init_groups_ini os-core test-bundle1 test-bundle2
+
+  set_os_release 10 os-core
+  track_bundle 10 os-core
+  track_bundle 10 test-bundle1
+  track_bundle 10 test-bundle2
+  gen_file_plain 10 test-bundle1 foo
+  gen_file_plain 10 test-bundle1 foobar
+  gen_file_plain 10 test-bundle2 foo2
+  gen_includes_file test-bundle2 10 test-bundle1
+
+  set_os_release 20 os-core
+  track_bundle 20 os-core
+  track_bundle 20 test-bundle1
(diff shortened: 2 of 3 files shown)


src/manifest.c
```diff
@@ -675,19 +675,21 @@ char *file_type_to_string(struct file *file)
 	return type;
 }
 
+/* Calculate the contentsize for the manifest based on file sizes.
+ * This should calculate the files uniquely included in this manifest, but none
+ * of its submanifests, which will allow calculation of all bundles on a system
```
phmccarty: instead of " ... calculation of all bundles ..." the above should say "... calculation of sizes of all bundles ...

src/manifest.c
```diff
@@ -675,19 +675,21 @@ char *file_type_to_string(struct file *file)
...
+ * by adding all manifest->contentsizes of installed bundles.
+ */
```
phmccarty: When a set of bundles are part of the same include hierarchy, the description of how clients can calculate total bundle size is true, but the description is not true in general.

For example, consider two identical bundles, A and B. Both bundles include bundle C. The contentsize of bundle C is subtracted from the size of bundles A and B to produce contentsize for A and B. However, if a client installs both bundles A and B, summing their contentsize values will be an overestimate, because they overlap 100%.

Can you clarify the doc string?

phmccarty (commented):
+1 on the changes. Can you add a new functional test as well?

src/manifest.c
matthewrsj: Ah, hadn't thought of that particular case. Yes I'll update.

gtkramer:
+1 as well, lgtm.  Thank you for working on this so quickly, @matthewrsj

src/manifest.c
phmccarty: Thanks.

Just to clarify in more detail: What I described is the worse case scenario. In reality, bundles not in the same include chain (however you want to describe this) often just partially overlap, content-wise. At that point, swupd-client would need to know content size at the *file level* instead of *manifest level* to determine the real content size. Because manifests do not list content size at the file level, determining the real content size is intractable with the current state of things.

icarus-sparry: Altering the manifest to add the size of each file has an additional advantage, in that "swupd verify --fast" could check the size (it is already doing a stat call so the only extra time is converting the file size from ascii in the manifest). It could also be used whilst trying to work out if an update can fit on the disk.

matthewrsj:
@phmccarty functional test added, docstring expanded.

test/functional/contentsize-across-versions-includes/test.bats
```diff
@@ -0,0 +1,57 @@
...
+@test "correct contentsize" {
+  # create a couple updates to both check that contentsize does not add included
+  # bundles and to verify that files changed in previous updates are counted.
+  sudo $CREATE_UPDATE --osversion 10 --statedir $DIR --format 3
+  set_latest_ver 10
+
+  # contentsize for test-bundle2 should not include test-bundle1's contentsize
+  [[ 1 -eq $(grep '^contentsize:	11$' $DIR/www/10/Manifest.test-bundle1 | wc -l) ]]
+  [[ 1 -eq $(grep '^contentsize:	5$'  $DIR/www/10/Manifest.test-bundle2 | wc -l) ]]
+  [[ 1 -eq $(grep '^contentsize:	5150$' $DIR/www/10/Manifest.full | wc -l) ]]
```
phmccarty: My first thought was "where does this large number come from?" :-) It is due to the fixed value used for directories (1024 bytes), and os-core includes 5 directories for this test.

Thus, I think it would be more clear to add an additional check here for the correct os-core contentsize. Then you can visually see that the Manifest.full size is the sum of all bundle sizes.

test/functional/contentsize-across-versions-includes/test.bats
```diff
@@ -0,0 +1,57 @@
...
+  sudo $CREATE_UPDATE --osversion 20 --statedir $DIR --format 3
+  set_latest_ver 20
+
+  # one new file: foobarbaz (10 bytes)
+  [[ 1 -eq $(grep '^contentsize:	21$' $DIR/www/20/Manifest.test-bundle1 | wc -l) ]]
+  # one new file: foo2bar (8 bytes)
+  [[ 1 -eq $(grep '^contentsize:	13$'  $DIR/www/20/Manifest.test-bundle2 | wc -l) ]]
+  # contentsize for full should be all files, including ones not changed in this release
+  # two new files: foo2bar (8 bytes) and foobarbaz (10 bytes)
+  [[ 1 -eq $(grep '^contentsize:	5168$' $DIR/www/20/Manifest.full | wc -l) ]]
```
phmccarty: again, check os-core contentsize here too

src/manifest.c
```diff
@@ -675,19 +675,29 @@ char *file_type_to_string(struct file *file)
...
+ * of its submanifests, which will allow calculation of sizes of all bundles on
+ * a system by adding all manifest->contentsizes of installed bundles.
+ *
+ * However, if two bundles not in the same include chain have overlapping
+ * content, summing the include chain of each bundle in the client will result
+ * in an over-estimation of the total size on the system. The more content is
+ * shared, the higher the over-estimation. In reality this overlap will not be
+ * large, but it is currently impossible to calculate the exact installed size
+ * using just the contentsize.
+ */
```
phmccarty, line 690: +1 on the comment revision. I would revise the commit description as well.

matthewrsj:
@phmccarty commit message updated, functional test updated.

phmccarty:
+1

PR #67, opened 2017-09-12
Prune debuginfo from manifests

matthewrsj:
If debuginfo under /usr/lib/debug/ or /usr/src/debug/ are added to the
manifests mistakenly, prune those files. Make sure manifest is rewritten
when debuginfo files are pruned, even if no new files are added.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -1032,12 +1032,13 @@ int remove_old_deleted_files(struct manifest *m1, struct manifest *m2)
  * Returns 0 when the pruned manifest no longer has new files.
  * Returns -1 on error.
  */
-int prune_manifest(struct manifest *manifest)
+int prune_manifest(struct manifest *manifest, bool *pruned)
 {
 	GList *list;
 	GList *next;
 	struct file *file;
 	int newfiles = 0;
+	*pruned = false;
 
 	/* prune some files */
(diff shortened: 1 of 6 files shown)


src/manifest.c
```diff
@@ -1055,6 +1056,10 @@ int prune_manifest(struct manifest *manifest)
 			// LOG(file, "Skipping deleted boot file in manifest write", "component %s", manifest->component);
 			manifest->files = g_list_delete_link(manifest->files, list);
 			manifest->count--;
+		} else if (is_debuginfo(file->filename)) {
+			manifest->files = g_list_delete_link(manifest->files, list);
+			manifest->count--;
+			*pruned = true;
```
phmccarty, line 1062: This approach makes sense for the case when a file with `/usr/lib/debug` is already present in a manifest (deleted or not) and we want to remove it later on. However, with this change in place, files with `/usr/lib/debug` would never be added to a manifest in the first place...

So, I'm not sure this is going to behave well, in practice, because manifests will be written whenever a debuginfo file is encountered, regardless if any other "changes" were made to the bundle.

matthewrsj:
@phmccarty you are correct, it would force manifest generation every time one of these files is detected. It is better to just ban the filename (make it 'illegal'). I'll open a different PR for that approach.

PR #70, opened 2017-09-15
Ban debuginfo from manifests

matthewrsj:
Debuginfo should not make it into the manifests since swupd-client is
unable to manage these files on client systems.


icarus-sparry:
Not totally happy about this. It is adding in more specialized clearlinux knowledge (that we mount fuse filesystems here). A more general solution might be to have a configuration file which has a list of banned paths.

icarus-sparry:
Much better.

phmccarty:
+1

PR #72, opened 2017-09-18
Prune debuginfo when configured to do so

### LINKED ISSUE #71: Deleted debuginfo still not removed from manifests
The latest change (0fe32ce8fbaabd19c55e8e92e8f0dfc7dbc48492) bans debuginfo during the chroot walk, but does not remove deletes from the manifests if they already exist. This will cause swupd-client to attempt to continue removing these files even though they are banned in swupd-server.

matthewrsj:
Fixes #71

Prune debuginfo from manifests when the [Debuginfo][pruned]
configuration is set to "true" in server.ini.


phmccarty:
+1

PR #81, opened 2017-09-27
Make full file creation for directories thread-safe

gtkramer:
When running many instances of swupd_make_fullfiles in parallel, the
read end of the pipe between the tar's for creating the full file of a
directory becomes reused.  The observed behavior of this is
swupd_make_fullfiles hangs indefinitely with the expected tar reader
missing.  The corresponding tar writer is not killed with SIGPIPE because
there is at least one reader still for the pipe, swupd_make_fullfiles.
First forking from swupd_make_fullfiles, creating the pipe, and then
fork-and-exec'ing for each tar ensures that pipe and file descriptor
management is contained for the directory rename in question and cannot
be reused by other directory renames.

Signed-off-by: George T Kramer <george.t.kramer@intel.com>

### DIFF
--- a/include/swupd.h
+++ b/include/swupd.h
@@ -262,10 +262,17 @@ extern FILE *fopen_exclusive(const char *filename); /* no mode, opens for write 
 extern void dump_file_info(struct file *file);
 extern void string_or_die(char **strp, const char *fmt, ...);
 extern void print_elapsed_time(const char *step, struct timeval *previous_time, struct timeval *current_time);
+extern int system_argv_pipe(char *const lhscmd[], char *const rhscmd[]);
+extern int system_argv_pipe_fd(int lnewstdinfd, int lnewstderrfd, char *const lhscmd[],
+			       int rnewstdoutfd, int rnewstderrfd, char *const rhscmd[]);
+extern void pipe_monitor(int lnewstdinfd, int lnewstderrfd, char *const lhscmd[],
+			 int rnewstdoutfd, int rnewstderrfd, char *const rhscmd[]);
 extern int system_argv(char *const argv[]);
-extern int system_argv_fd(char *const argv[], int newstdin, int newstdout, int newstderr);
-extern int system_argv_pipe(char *const argvp1[], int stdinp1, int stderrp1,
-			    char *const argvp2[], int stdoutp2, int stderrp2);
+extern int system_argv_fd(int newstdinfd, int newstdoutfd, int newstderrfd, char *const cmd[]);
+extern pid_t system_argv_fd_nowait(int newstdinfd, int newstdoutfd, int newstderrfd, int closefd, char *const cmd[]);
+extern void exec_cmd_fd(int newstdinfd, int newstdoutfd, int newstderrfd, int closefd, char *const cmd[]);
+extern void move_fd(int oldfd, int newfd);
+extern int wait_process_terminate(pid_t pid);
 extern int num_threads(float scaling);
 extern bool file_is_debuginfo(const char *path);
 
--- a/src/fullfiles.c
+++ b/src/fullfiles.c
@@ -50,7 +50,6 @@ static void create_fullfile(struct file *file)
 	struct stat sbuf;
 	char *empty, *indir, *outdir;
 	char *param1, *param2;
-	int stderrfd;
 
 	if (file->is_deleted) {
 		return; /* file got deleted -> by definition we cannot tar it up */
@@ -98,17 +97,13 @@ static void create_fullfile(struct file *file)
 		char *const tarcfcmd[] = { TAR_COMMAND, "-C", dir, TAR_PERM_ATTR_ARGS_STRLIST, "-cf", "-", param1, param2, NULL };
 		char *const tarxfcmd[] = { TAR_COMMAND, "-C", rename_tmpdir, TAR_PERM_ATTR_ARGS_STRLIST, "-xf", "-", NULL };
 
-		stderrfd = open("/dev/null", O_WRONLY);
-		if (stderrfd == -1) {
-			LOG(NULL, "Failed to open /dev/null", "");
-			assert(0);
-		}
-		if (system_argv_pipe(tarcfcmd, -1, stderrfd, tarxfcmd, -1, stderrfd) != 0) {
+		int tarcmdresult = system_argv_pipe(tarcfcmd, tarxfcmd);
+		if (tarcmdresult != 0) {
+			LOG(NULL, "Tar command for copying directory full file failed with code %d", tarcmdresult);
 			assert(0);
 		}
 		free(param1);
 		free(param2);
-		close(stderrfd);
 
 		string_or_die(&rename_source, "%s/%s", rename_tmpdir, base);
 		string_or_die(&rename_target, "%s/%s", rename_tmpdir, file->hash);
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -149,130 +149,117 @@ void concat_str_array(char **output, char *const argv[])
 	}
 }
 
+int system_argv_pipe(char *const lhscmd[], char *const rhscmd[])
+{
+	return system_argv_pipe_fd(-1, -1, lhscmd, -1, -1, rhscmd);
+}
+
+int system_argv_pipe_fd(int lnewstdinfd, int lnewstderrfd, char *const lhscmd[],
+			int rnewstdoutfd, int rnewstderrfd, char *const rhscmd[])
+{
+	pid_t monitorpid = fork();
+	if (monitorpid == -1) {
+		LOG(NULL, "Failed to create child process to monitor pipe between", "command %s and command %s", lhscmd[0], rhscmd[0]);
+		return -1;
+	} else if (monitorpid == 0) {
+		pipe_monitor(lnewstdinfd, lnewstderrfd, lhscmd, rnewstdoutfd, rnewstderrfd, rhscmd);
+	}
+	return wait_process_terminate(monitorpid);
+}
+
+void pipe_monitor(int lnewstdinfd, int lnewstderrfd, char *const lhscmd[],
+		  int rnewstdoutfd, int rnewstderrfd, char *const rhscmd[])
+{
+	int pipefd[2];
+	if (pipe(pipefd) == -1) {
+		LOG(NULL, "Failed to create a pipe between", "command %s and command %s", lhscmd[0], rhscmd[0]);
+		assert(0);
+	}
+
+	pid_t lhspid = system_argv_fd_nowait(lnewstdinfd, pipefd[1], lnewstderrfd, pipefd[0], lhscmd);
+	pid_t rhspid = system_argv_fd_nowait(pipefd[0], rnewstdoutfd, rnewstderrfd, pipefd[1], rhscmd);
+
+	if (close(pipefd[1]) == -1) {
+		LOG(NULL, "Could not close write end of pipe file descriptor", "%d", pipefd[1]);
+		assert(0);
+	}
+	if (close(pipefd[0]) == -1) {
+		LOG(NULL, "Could not close read end of pipe file descriptor", "%d", pipefd[0]);
+		assert(0);
+	}
+
+	int lhsresult = wait_process_terminate(lhspid);
+	int rhsresult = wait_process_terminate(rhspid);
+	exit(rhsresult != EXIT_SUCCESS ? rhsresult : lhsresult);
+}
+
 int system_argv(char *const argv[])
 {
-	int child_exit_status;
-	pid_t pid;
-	int status = -1;
-
-	pid = fork();
-
-	if (pid == 0) { /* child */
-		execvp(*argv, argv);
-		LOG(NULL, "This line must not be reached", "");
-		assert(0);
-	} else if (pid < 0) {
-		LOG(NULL, "Failed to fork a child process", "");
-		assert(0);
-	} else {
-		pid_t ws = waitpid(pid, &child_exit_status, 0);
-
-		if (ws == -1) {
-			LOG(NULL, "Failed to wait for child process", "");
-			assert(0);
-		}
-
-		if (WIFEXITED(child_exit_status)) {
-			status = WEXITSTATUS(child_exit_status);
-		} else {
-			LOG(NULL, "Child process didn't exit", "");
-			assert(0);
-		}
-
-		if (status != 0) {
-			char *cmdline = NULL;
-
-			concat_str_array(&cmdline, argv);
-			LOG(NULL, "Failed to run command:", "%s", cmdline);
-			free(cmdline);
-		}
-	}
-
-	return status;
+	return system_argv_fd(-1, -1, -1, argv);
 }
 
-int system_argv_fd(char *const argv[], int newstdin, int newstdout, int newstderr)
+int system_argv_fd(int newstdinfd, int newstdoutfd, int newstderrfd, char *const cmd[])
 {
-	int child_exit_status;
-	pid_t pid;
-	int status = -1;
-
-	pid = fork();
-
-	if (pid == 0) { /* child */
-		if (newstdin >= 0) {
-			if (dup2(newstdin, STDIN_FILENO) == -1) {
-				LOG(NULL, "Could not redirect stdin", "");
-				assert(0);
-			}
-			close(newstdin);
-		}
-		if (newstdout >= 0) {
-			if (dup2(newstdout, STDOUT_FILENO) == -1) {
-				LOG(NULL, "Could not redirect stdout", "");
-				assert(0);
-			}
-			close(newstdout);
-		}
-		if (newstderr >= 0) {
-			if (dup2(newstderr, STDERR_FILENO) == -1) {
-				LOG(NULL, "Could not redirect stderr", "");
-				assert(0);
-			}
-			close(newstderr);
-		}
-
-		execvp(*argv, argv);
-		LOG(NULL, "This line must not be reached", "");
-		assert(0);
-	} else if (pid < 0) {
-		LOG(NULL, "Failed to fork a child process", "");
-		assert(0);
-	} else {
-		pid_t ws = waitpid(pid, &child_exit_status, 0);
-
-		if (ws == -1) {
-			LOG(NULL, "Failed to wait for child process", "");
-			assert(0);
-		}
-
-		if (WIFEXITED(child_exit_status)) {
-			status = WEXITSTATUS(child_exit_status);
-		} else {
-			LOG(NULL, "Child process didn't exit", "");
-			assert(0);
-		}
-
-		if (status != 0) {
-			char *cmdline = NULL;
-
-			concat_str_array(&cmdline, argv);
-			LOG(NULL, "Failed to run command:", "%s", cmdline);
-			free(cmdline);
-		}
-	}
-
-	return status;
+	pid_t cmdpid = system_argv_fd_nowait(newstdinfd, newstdoutfd, newstderrfd, -1, cmd);
+	return wait_process_terminate(cmdpid);
 }
 
-int system_argv_pipe(char *const argvp1[], int stdinp1, int stderrp1,
-		     char *const argvp2[], int stdoutp2, int stderrp2)
+pid_t system_argv_fd_nowait(int newstdinfd, int newstdoutfd, int newstderrfd, int closefd, char *const cmd[])
 {
-	int statusp2;
-	int pipefd[2];
-
-	if (pipe(pipefd)) {
-		LOG(NULL, "Failed to create a pipe", "");
-		return -1;
+	pid_t cmdpid = fork();
+	if (cmdpid == -1) {
+		LOG(NULL, "Failed to fork to execute command", "%s", cmd[0]);
+		assert(0);
+	} else if (cmdpid == 0) {
+		exec_cmd_fd(newstdinfd, newstdoutfd, newstderrfd, closefd, cmd);
 	}
-	system_argv_fd(argvp1, stdinp1, pipefd[1], stderrp1);
-	close(pipefd[1]);
-	statusp2 = system_argv_fd(argvp2, pipefd[0], stdoutp2, stderrp2);
-	close(pipefd[0]);
+	return cmdpid;
+}
 
-	/* Returns the status of the failed process if any
-       If both processes failed returns the status of first one */
-	return statusp2;
+void exec_cmd_fd(int newstdinfd, int newstdoutfd, int newstderrfd, int closefd, char *const cmd[])
+{
+	move_fd(newstdinfd, STDIN_FILENO);
+	move_fd(newstdoutfd, STDOUT_FILENO);
+	move_fd(newstderrfd, STDERR_FILENO);
+	if (closefd >= 0 && close(closefd) == -1) {
+		LOG(NULL, "Could not close file descriptor", "%d", closefd);
+		assert(0);
+	}
+	execvp(*cmd, cmd);
+	LOG(NULL, "Command", "%s failed", cmd[0]);
+	assert(0);
+}
+
+void move_fd(int oldfd, int newfd)
+{
+	if (oldfd < 0 || newfd < 0 || oldfd == newfd) {
+		return;
+	}
+	if (dup2(oldfd, newfd) == -1) {
+		LOG(NULL, "Could not create duplicate file descriptor", "%d from %d", newfd, oldfd);
+		assert(0);
+	}
+	if (close(oldfd) == -1) {
+		LOG(NULL, "Could not close file descriptor", "%d", oldfd);
+		assert(0);
+	}
+}
+
+int wait_process_terminate(pid_t pid)
+{
+	int status;
+	do {
+		if (waitpid(pid, &status, 0) == -1) {
+			LOG(NULL, "Failed to wait for PID", "%d", pid);
+			return -1;
+		}
+	} while (!WIFEXITED(status) && !WIFSIGNALED(status));
+	// Exit statuses fall in the range of [0, 255].  Make signal statuses fall in a non-overlapping range starting with 256.
+	if (WIFEXITED(status)) {
+		return WEXITSTATUS(status);
+	} else {
+		return 256 + WTERMSIG(status);
+	}
 }
 
 void check_root(void)


src/helpers.c
```diff
@@ -191,88 +191,80 @@ int system_argv(char *const argv[])
...
-			if (dup2(newstdin, STDIN_FILENO) == -1) {
-				LOG(NULL, "Could not redirect stdin", "");
-				assert(0);
-			}
-			close(newstdin);
+	pid_t childpid = fork();
+	if (childpid == -1) {
+		LOG(NULL, "Failed to create child process", "");
+		return -1;
+	} else if (childpid == 0) {
```
icarus-sparry: As a matter of style, I would not have an "else" here, as the end of the "if" finishes with an unconditional change of flow (return in this case, but other places have break and continue). If we do this we gain back a level of indentation.

You could go even further, and create tiny functions like 'pipeline_monitor', 'lhs' and 'rhs' so the code becomes

pid_t pipelinemonpid = fork();
if (pipelinemonpid == -1) {
    LOG(...)
   return -1;
}
if pipelinemonpid == 0) {
     pipeline_monitor(lhscmd, rhscmd)
}
return wait_process_terminate(pipelinemonpid);

I like small functions!

src/helpers.c
```diff
@@ -191,88 +191,80 @@ int system_argv(char *const argv[])
...
+			assert(0);
 		}
-		if (newstdout >= 0) {
-			if (dup2(newstdout, STDOUT_FILENO) == -1) {
+
+		pid_t lhspid = fork();
+		if (lhspid == -1) {
+			LOG(NULL, "Failed to fork LHS process", "");
+			return -1;
+		} else if (lhspid == 0) {
```
icarus-sparry: And gain a level of indentation here, even if you don't put this in a separate function.

src/helpers.c
```diff
@@ -191,88 +191,80 @@ int system_argv(char *const argv[])
...
+			execvp(*lhscmd, lhscmd);
+			LOG(NULL, "LHS command failed", "");
+			assert(0);
+		} else {
+			pid_t rhspid = fork();
+			if (rhspid == -1) {
+				LOG(NULL, "Failed to fork RHS process", "");
+				return -1;
+			} else if (rhspid == 0) {
+				if (dup2(pipefd[0], STDIN_FILENO) == -1) {
```
icarus-sparry: Really here you want to move the file descriptor. This means you want to close pipefd[0] unless it value is equal to STDIN_FILENO.

src/helpers.c
```diff
@@ -191,88 +191,80 @@ int system_argv(char *const argv[])
...
+					LOG(NULL, "Could not redirect stdin", "");
+					assert(0);
+				}
+				execvp(*rhscmd, rhscmd);
+				LOG(NULL, "LHS command failed", "");
```
icarus-sparry: LHS -> RHS

src/helpers.c
```diff
@@ -191,88 +191,80 @@ int system_argv(char *const argv[])
...
+			if (dup2(pipefd[1], STDOUT_FILENO) == -1) {
 				LOG(NULL, "Could not redirect stdout", "");
 				assert(0);
 			}
-			close(newstdout);
-		}
-		if (newstderr >= 0) {
-			if (dup2(newstderr, STDERR_FILENO) == -1) {
-				LOG(NULL, "Could not redirect stderr", "");
+			execvp(*lhscmd, lhscmd);
```
icarus-sparry: Missing important code here, see comments below.

src/helpers.c
```diff
@@ -191,88 +191,80 @@ int system_argv(char *const argv[])
...
+			pid_t rhspid = fork();
+			if (rhspid == -1) {
+				LOG(NULL, "Failed to fork RHS process", "");
+				return -1;
+			} else if (rhspid == 0) {
+				if (dup2(pipefd[0], STDIN_FILENO) == -1) {
+					LOG(NULL, "Could not redirect stdin", "");
+					assert(0);
+				}
+				execvp(*rhscmd, rhscmd);
```
icarus-sparry: Missing important code here as well, see comments below.

src/helpers.c
```diff
@@ -191,88 +191,80 @@ int system_argv(char *const argv[])
...
+				LOG(NULL, "LHS command failed", "");
 				assert(0);
+			} else {
+				int lhsresult = wait_process_terminate(lhspid);
+				int rhsresult = wait_process_terminate(rhspid);
+				close(pipefd[0]);
+				close(pipefd[1]);
+				if (lhsresult == 0 && rhsresult == 0) {
```
icarus-sparry: I would just code this as

exit(rhsresult != EXIT_SUCCESS? rhsresult : lhsresult);

src/helpers.c
```diff
@@ -191,88 +191,80 @@ int system_argv(char *const argv[])
...
-		}
-
-		if (status != 0) {
-			char *cmdline = NULL;
-
-			concat_str_array(&cmdline, argv);
-			LOG(NULL, "Failed to run command:", "%s", cmdline);
-			free(cmdline);
-		}
+		int childresult = wait_process_terminate(childpid);
```
icarus-sparry: Again get back a level of indentation by not using else.

src/helpers.c
```diff
@@ -191,88 +191,80 @@ int system_argv(char *const argv[])
...
-		LOG(NULL, "Failed to create a pipe", "");
-		return -1;
+	int status;
+	do {
+		if (waitpid(pid, &status, 0) == -1) {
+			LOG(NULL, "Failed to wait for PID", "");
+			return -1;
+		}
+	} while (!WIFEXITED(status) && !WIFSIGNALED(status));
+	if (WIFEXITED(status)) {
```
icarus-sparry: Not a great fan of this return code stuff. It makes it hard to tell if tar returned say 6 or tar was killed with a SIGABRT.

Either just return the status value as is, or do what ksh does and return 256+WTERMSIG(status) in the signal case. (Note bash retuns 128+WTERMSIG(status) which doesn't allow you to distinguish half the valid return codes, but is better than nothing).

include/swupd.h
```diff
@@ -263,9 +263,8 @@ extern void dump_file_info(struct file *file);
 extern void string_or_die(char **strp, const char *fmt, ...);
 extern void print_elapsed_time(const char *step, struct timeval *previous_time, struct timeval *current_time);
 extern int system_argv(char *const argv[]);
-extern int system_argv_fd(char *const argv[], int newstdin, int newstdout, int newstderr);
-extern int system_argv_pipe(char *const argvp1[], int stdinp1, int stderrp1,
-			    char *const argvp2[], int stdoutp2, int stderrp2);
+extern int system_argv_pipe(char *const lhscmd[], char *const rhscmd[]);
```
icarus-sparry: I quite liked the old api, but if no one was every using it then the slimmed down one is better.

src/fullfiles.c
```diff
@@ -98,17 +97,11 @@ static void create_fullfile(struct file *file)
 		char *const tarcfcmd[] = { TAR_COMMAND, "-C", dir, TAR_PERM_ATTR_ARGS_STRLIST, "-cf", "-", param1, param2, NULL };
 		char *const tarxfcmd[] = { TAR_COMMAND, "-C", rename_tmpdir, TAR_PERM_ATTR_ARGS_STRLIST, "-xf", "-", NULL };
 
-		stderrfd = open("/dev/null", O_WRONLY);
-		if (stderrfd == -1) {
-			LOG(NULL, "Failed to open /dev/null", "");
-			assert(0);
-		}
-		if (system_argv_pipe(tarcfcmd, -1, stderrfd, tarxfcmd, -1, stderrfd) != 0) {
+		if (system_argv_pipe(tarcfcmd, tarxfcmd) != 0) {
```
icarus-sparry: We go to all the trouble to get a return code, we should log it before aborting.

matthewrsj:
Please rebase to latest master to fix CI error.

src/helpers.c
```diff
@@ -191,88 +191,80 @@ int system_argv(char *const argv[])
...
-		if (newstdout >= 0) {
-			if (dup2(newstdout, STDOUT_FILENO) == -1) {
+
+		pid_t lhspid = fork();
+		if (lhspid == -1) {
+			LOG(NULL, "Failed to fork LHS process", "");
+			return -1;
+		} else if (lhspid == 0) {
+			if (dup2(pipefd[1], STDOUT_FILENO) == -1) {
 				LOG(NULL, "Could not redirect stdout", "");
```
icarus-sparry: Again you want to move the fd, so close it after the dup2 unless it is STDOUT_FILENO

icarus-sparry (changes requested):

src/helpers.c
gtkramer: Say there's a lot of data that needs to move between the lhs tar and the rhs tar.  Does waiting in the pipe monitor function prevent the rhs tar from being called in this case because the wait_process_terminate function does not acknowledge if a process was stopped or continued?  Because the lhs tar would have blocked once the pipe buffer would have filled up, waiting for the rhs tar to consume data from the pipe, right?  But the rhs tar would never be exec'd because we'd be waiting on the lhs tar?

icarus-sparry: The 3 processes are able to run independently until they interact. The interactions are the pipe between lhs and rhs, where the lhs will pause until the rhs has read the data, and between the pipe monitor and the two processes it creates when it calls wait.

**The wait(2) call says 'put the pipemonitor to sleep until the lhs (or rhs, depending on what you are waiting for) has a change of state (essentially finished or get a signal)'.**

So the amount of data being transferred between the lhs and rhs might make a difference to the amount of clock time that passes before the pipe monitor is awoken, but nothing else.

The wait_process_terminate is ignores some state changes, although it could ignore more.

src/helpers.c
gtkramer: Ah, even the close(2) has an obvious warning:
"Since a file descriptor may be reused, there are some obscure race conditions that may cause unintended side effects."

src/helpers.c
gtkramer: Ah, ok.  I'm compiling and testing changes which incorporate all of your feedback and bring back the interface for managing file descriptors with commands.  If things go well, I'll add another commit which unifies all system_argv commands under common code and interfaces using small functions.

src/helpers.c
icarus-sparry: Sorry, the "important code" which is missing here is something like
close_if_not_std(pipefd[0]); close_if_not_std(pipefd[1]);close_if_not_std(newstderr);

and add

static void close_if_not_std(int fd){
switch (fd) {
case STDIN_FILENO:
case STDOUT_FILENO:
case STDERR_FILENO:
   break;
default:
   close(fd);/* TODO, should this log failures? return them? */
}
}

src/helpers.c
icarus-sparry: See close_if_not_std(pipefd[0) above

src/helpers.c
```diff
@@ -191,88 +191,80 @@ int system_argv(char *const argv[])
...
+			} else if (rhspid == 0) {
+				if (dup2(pipefd[0], STDIN_FILENO) == -1) {
+					LOG(NULL, "Could not redirect stdin", "");
+					assert(0);
+				}
+				execvp(*rhscmd, rhscmd);
+				LOG(NULL, "LHS command failed", "");
 				assert(0);
+			} else {
+				int lhsresult = wait_process_terminate(lhspid);
```
icarus-sparry: It is **critical** that the 2 close calls are done before the 2 wait_process_terminate calls, to avoid deadlocks.

gtkramer:
The branch was rebased and force updated.  It includes all of the changes suggested from @icarus-sparry and team from earlier discussion.  It has been confirmed to fix the hang we were seeing as well, and make clean checks out.

src/helpers.c
```diff
@@ -149,130 +149,108 @@ void concat_str_array(char **output, char *const argv[])
...
-
-			concat_str_array(&cmdline, argv);
-			LOG(NULL, "Failed to run command:", "%s", cmdline);
-			free(cmdline);
+		pid_t rhspid = system_argv_fd_nowait(pipefd[0], -1, -1, rhscmd);
+		if (close(pipefd[0]) == -1) {
+			LOG(NULL, "Could not close read end of pipe file descriptor", "%d", pipefd[0]);
+			assert(0);
 		}
+
```
icarus-sparry: You **must** close both pipefd[0] and pipefd[1] between lines 180 and 181. If you don't then if the RHS for example exits for some reason, but the LHS does not then the LHS will never get a SIGPIPE, but will just block in the write to the pipe and you will get a deadlock. Likewise if the LHS exists for some reason, the RHS will block in the read call as it will not get EOF. This is because the process which is running the code in lines 181 to 183 has the file descriptors open.

src/helpers.c
```diff
@@ -149,130 +149,108 @@ void concat_str_array(char **output, char *const argv[])
...
-		     char *const argvp2[], int stdoutp2, int stderrp2)
+void move_fd(int oldfd, int newfd)
 {
-	int statusp2;
-	int pipefd[2];
+	if (dup2(oldfd, newfd) == -1) {
+		LOG(NULL, "Could not create duplicate file descriptor", "%d from %d", newfd, oldfd);
+		assert(0);
+	}
+	if (close(oldfd) == -1) {
```
icarus-sparry: This is dangerous. Consider what happens if someone wants to say combine stdout and stderr and send them both down the pipe (i.e. "tar xvf fred |& tee /tmp/log"). Or what happens if oldfd and newfd are the same.

At the very least you need to just return doing nothing if they are the same.

src/helpers.c
```diff
@@ -149,130 +149,108 @@ void concat_str_array(char **output, char *const argv[])
...
+int wait_process_terminate(pid_t pid)
+{
+	int status;
+	do {
+		if (waitpid(pid, &status, 0) == -1) {
+			LOG(NULL, "Failed to wait for PID", "%d", pid);
+			return -1;
+		}
+	} while (!WIFEXITED(status) && !WIFSIGNALED(status));
+	// Exit statuses fall in the range of [0, 255].  Make signal statuses fall in a non-overlapping range starting with 256 (128 + (128 + N)).
```
icarus-sparry: Comment doesn't match code. WTERMSIG returns values in [0, 127], so you can't tell if 130 mean exit(130) or signal 2.

src/helpers.c
```diff
@@ -149,130 +149,108 @@ void concat_str_array(char **output, char *const argv[])
...
+	}
 }
 
-int system_argv_pipe(char *const argvp1[], int stdinp1, int stderrp1,
-		     char *const argvp2[], int stdoutp2, int stderrp2)
+void move_fd(int oldfd, int newfd)
 {
-	int statusp2;
-	int pipefd[2];
+	if (dup2(oldfd, newfd) == -1) {
```
icarus-sparry: I would put the check in here for oldfd being != -1, and remove lines 216, 218, 219, 221, 222, and 224, but I don't feel strongly about it.

src/helpers.c
```diff
@@ -149,130 +149,108 @@ void concat_str_array(char **output, char *const argv[])
...
+		if (pipe(pipefd) == -1) {
+			LOG(NULL, "Failed to create a pipe between", "command %s and command %s", lhscmd[0], rhscmd[0]);
 			assert(0);
 		}
-
-		if (WIFEXITED(child_exit_status)) {
-			status = WEXITSTATUS(child_exit_status);
-		} else {
-			LOG(NULL, "Child process didn't exit", "");
+		pid_t lhspid = system_argv_fd_nowait(-1, pipefd[1], -1 , lhscmd);
```
icarus-sparry: I think you want
pid_t lhspid = system_argv_fd_nowait(lnewstdinfd, pipefd[1], lnewstderrfd , lhscmd); 
and a similar change at line 175. 

Note comment on line 206 means you need to pass both pipefd[0] and pipefd[1] here, and on line 175 pass pipfd[1] and pipefd[0] (i.e. reversed)

icarus-sparry (changes requested):
Sorry, the file descriptor manipulation is still not right and can lead to deadlocks.

src/helpers.c
gtkramer: pipefd[0] and pipefd[1] are being closed on lines 176 and 171, respectively, after their corresponding fork-execs happen but before the waiting happens.  This is the correct place to close them, right?

src/helpers.c
gtkramer: Oops, yes, I forgot to change this.  Thank you for catching this!

src/helpers.c
gtkramer: That makes sense. It's a good way to reduce the code duplication!

src/helpers.c
gtkramer: I misread the documentation here.  This error handling method would only work to properly spread out fatal error signal N, which is coded 128 + N.  It would not work properly for all error signals coded in the range [0, 255]: http://tldp.org/LDP/abs/html/exitcodes.html

src/helpers.c
gtkramer: I might be misunderstanding the second part of the comment.  Passing both ends of the pipe on lines 170 and 175 would, in addition to the current behavior, set the STDIN for the LHS command to be the read end of the pipe, and set the STDOUT for the RHS command to be the write end of the pipe.  I didn't set these because I didn't think commands involved in a pipe should be reading or writing data in this direction; creating something like a loop for data flow within each command.  If a chain of piped commands was being made where additional redirection was needed, another pipe should be made, right?

src/helpers.c
gtkramer: Oh, yes, I see.  The code assumes these to be different.  It doesn't handle the case where you really just want to symbolically link the file descriptors *and* keep them around.  I'll see if I can come up with a better revision for this.  Otherwise, I'll put in a check to ensure they are different and log when they are not.

src/helpers.c
icarus-sparry: Ah, sorry. I was mismatching the "}" on line 179. As I mentioned before I find the code to have too much in a single function. So in the pipemonitor process you have the needed closes at lines 171 and 176. You are missing the needed closes in the lhs and rhs processes.

src/helpers.c
```diff
@@ -149,130 +149,108 @@ void concat_str_array(char **output, char *const argv[])
...
+pid_t system_argv_fd_nowait(int newstdinfd, int newstdoutfd, int newstderrfd, char *const cmd[])
+{
+	pid_t cmdpid = fork();
+	if (cmdpid == -1) {
+		LOG(NULL, "Failed to fork to execute command", "%s", cmd[0]);
 		assert(0);
-	} else if (pid < 0) {
-		LOG(NULL, "Failed to fork a child process", "");
+	} else if (cmdpid == 0) {
+		set_std_fd(newstdinfd, newstdoutfd, newstderrfd);
```
icarus-sparry: It is **critical** that the other half of the pipe is closed between lines 205 and 207 to avoid deadlocks. This probably means you need to pass in an extra parameter of a file descriptor to be closed, and call close if it is not -1. 

If you don't do this, then you can have something like the following. Say the pipe call returns 21 and 22 at line 166. This is in the pipemonitor process. You then call system_argv_fd_nowait at line 170, which creates the lhs process, which also has 21 and 22 open. 22 is then moved to stdout, but 21 is left open, so this process has both the read and the write ends of the pipe open on fd 1 and 21.
Back in the pipemonitor process at line 171 you close 22.
At line 175 you call system_argv_fd_nowait to create the rhs. So this also has 21 open. This is moved to stdin in the rhs, and closed at line 176 in pipemonitor.
So if the rhs dies for some reason, the lhs is left with both file descriptors to the pipe open, and so will never receive sigpipe but will just block in the write call.

So I think you need to pass an extra parameter to system_argv_fd_nowait which is a fd to close if it is non zero. You need to pass it as you want to close it only in the child around line 206.

I would make the code rather more symmetric, I would create the two processes first and then close the file descriptors in the pipemonitor process. In the system_argv_fd_nowait I would close the unneeded fd in the child before the exec.

Another way to do this would be to replace the pipe call on line 166 with pipe2(pipefd, O_CLOEXEC) to make the execvp close the unneeded descriptors.

src/helpers.c
icarus-sparry: I think system_argv_fd_nowait requires a 5th parameter, a fd to optionally close. See comment on line 206.

(I think I am seeing how I am driving GitHub reviews incorrectly. I have multiple comments open and then press "finish review" and it doesn't warn me that these comments have not been saved).

icarus-sparry (changes requested):

src/helpers.c
gtkramer: If I wanted to summarize our pipe file descriptor management discussions, the long and short of it is: do not keep unnecessary pipe file descriptors open in any process, forked or not, because otherwise it introduces the risk of accidental reuse.  We want a guarantee that this won't happen, so we close pipe file descriptors immediately if they are not needed before exec'ing any commands.  Being lax about pipe file descriptor management can block SIGPIPE from being delivered, or from EOF being given to a command; leading to hung processes.  Does that sound about right?

(That sounds like it was how the large comment was missed during our review on Wednesday.)

src/helpers.c
gtkramer: Ah, ok.  My first pass broke chunky functions down into smaller one, but I should probably keep going.

src/helpers.c
gtkramer: Roger that.  I'm working on the changes for this, and adding the additional parameter to system_argv_fd_nowait seems to work well.

In my first stab at the second revision for this, before I committed, I had the code symmetric, but I had moved the first close up so that pipe file descriptor was closed a bit earlier.  However, the symmetry makes the code easier to read.  I'll move it back.

src/helpers.c
icarus-sparry: I would not log if they are the same, for 2 reasons. 1. The message is not actionable, you can't do anything if the message appears, and 2. it is not a bug. For example, if you ran 'make_fullfiles >&-' which closes stdout, assuming that the program made it as far as here then one of the file descriptors being returned by the pipe call could well be 1. Then you would want to move 1 to 1, which is a do nothing operation, not a bug.

src/helpers.c
icarus-sparry: Yes, that is a good summary.

https://www.youtube.com/watch?v=JDSPwexlyTo

gtkramer:
The branch has been updated to include all of the recent changes suggested from @icarus-sparry.  As before, the recent changes have been confirmed to fix the hang we were seeing.  The checks for "make clean" and "make compliant" have been run and look okay.

src/helpers.c
```diff
@@ -149,130 +149,116 @@ void concat_str_array(char **output, char *const argv[])
...
-		LOG(NULL, "This line must not be reached", "");
-		assert(0);
-	} else if (pid < 0) {
-		LOG(NULL, "Failed to fork a child process", "");
+void pipe_monitor(int lnewstdinfd, int lnewstderrfd, char *const lhscmd[],
+		  int rnewstdoutfd, int rnewstderrfd, char *const rhscmd[])
+{
+	int pipefd[2];
+	if (pipe(pipefd) == -1) {
+		LOG(NULL, "Failed to create a pipe between", "command %s and command %s", lhscmd[0], rhscmd[0]);
```
icarus-sparry, line 175: The comma between the 2 constant strings looks wrong. Without actually looking I would expect the second argument to LOG to be a printf style string.  If there is not a compile time warning we should update our headers to get it to do so.

src/helpers.c
```diff
@@ -149,130 +149,116 @@ void concat_str_array(char **output, char *const argv[])
...
-			status = WEXITSTATUS(child_exit_status);
-		} else {
-			LOG(NULL, "Child process didn't exit", "");
-			assert(0);
-		}
+	}
 
-		if (status != 0) {
-			char *cmdline = NULL;
+	pid_t lhspid = system_argv_fd_nowait(lnewstdinfd, pipefd[1], lnewstderrfd, &pipefd[0], lhscmd);
```
icarus-sparry: This works, but I would have coded system_argv_fd_nowait to have its 4th parameter as an int, rather than a pointer to int, and used -1 to indicate there is nothing to close. Passing pointers around makes it harder to reason about code.

src/helpers.c
```diff
@@ -149,130 +149,116 @@ void concat_str_array(char **output, char *const argv[])
...
+	execvp(*cmd, cmd);
+	LOG(NULL, "Command", "%s failed", cmd[0]);
+	assert(0);
+}
 
-		if (ws == -1) {
-			LOG(NULL, "Failed to wait for child process", "");
+void move_fd(int oldfd, int newfd)
+{
+	if (oldfd >= 0 && newfd >= 0 && oldfd != newfd) {
```
icarus-sparry: Again this works, but I would invert the logic to "if (oldfd < 0 || newfd < 0 || oldfd == newfd ) { return ; }

icarus-sparry (changes requested):

src/helpers.c
gtkramer: LOG is used in both ways in the code base.  I too think it makes more sense to model this after printf usage than with this alternative usage.  Here's the definition of LOG:

`#define LOG(file, msg, fmt...) __log_message(file, msg, __FILE__, __LINE__, fmt)`

I think we can safely remove the extra comma?

src/helpers.c
gtkramer: Good point.  I'll fix this :)

src/helpers.c
gtkramer: This will let us decrease indentation as well ;)  I'll also fix this.

src/helpers.c
icarus-sparry: Ah, OK, if that is the definition then your code was correct, and you should ignore me.

gtkramer: OK, I'll go with the extra comma usage then.

icarus-sparry:
+1 from me, now George has explained LOG.

gtkramer:
The branch has been updated to include all of the recent changes suggested from @icarus-sparry. As before, the recent changes have been confirmed to fix the hang we were seeing. The checks for "make clean" and "make compliant" have been run and look okay.

src/helpers.c
```diff
@@ -149,130 +149,117 @@ void concat_str_array(char **output, char *const argv[])
...
+		assert(0);
+	}
 
-		if (ws == -1) {
-			LOG(NULL, "Failed to wait for child process", "");
-			assert(0);
-		}
+	int lhsresult = wait_process_terminate(lhspid);
+	int rhsresult = wait_process_terminate(rhspid);
+	exit(rhsresult != EXIT_SUCCESS ? rhsresult : lhsresult);
```
phmccarty, line 193: Should we return `rhsresult` unconditionally here, which is (IIUC) the default behavior of the Bash, unless the `pipefail` option is set? Even then, `pipefail` equivalent behavior could be coded as:

```
exit(lhsresult != EXIT_SUCCESS ? lhsresult : rhsresult);
```

icarus-sparry: Your understanding is perfectly correct. For this particular case knowing the fate of the rhs is more useful than knowing the fate of the lhs (which will probably die with sigpipe if the rhs exits). However knowing if either operation fails, e.g. the lhs is unable to read a file is also useful. For these reasons I think that we don't want the default bash behaviour, and we do want to return the exit status of the rhs if it is non-zero. For a rewrite in bash, we should look at the PIPESTATUS array.

phmccarty: Okay, that sounds reasonable to me. Let's keep this code as-is.

phmccarty:
+1 from me, @gtkramer

matthewrsj:
+1 here too

PR #83, opened 2017-09-28
Remove timestamp field from manifests

### LINKED ISSUE #79: Remove manifest timestamp
From https://github.com/clearlinux/swupd-client/issues/95:

We've never used this timestamp for anything. Manifests are meant to be immutable once created. So the timestamps should only flow forward and no issues.

In the unfortunate situation of a DevOps group needing to recreate old Manifests, they will have a new timestamp, and a changed hash relative what an old MoM (ie: assume MoM's got recreated, but some clients had downloaded and cached an old MoM, then downloaded a new bundle manifest on bundle  add), which leads to an error/warn in the client at runtime.

The error/warn can be addressed via https://github.com/clearlinux/swupd-client/issues/94, but that code path would also have never hit if the manifest did not have the timestamp in it.

Let's simply remove the timestamp. This can be done without a format bump.

matthewrsj:
Fixes #79

The timestamp field is not used and is a barrier to reproducible builds.
Remove the timestamp field from the manifests.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -751,7 +751,6 @@ static int write_manifest_plain(struct manifest *manifest)
 	fprintf(out, "version:\t%i\n", manifest->version);
 	fprintf(out, "previous:\t%i\n", manifest->prevversion);
 	fprintf(out, "filecount:\t%i\n", manifest->count);
-	fprintf(out, "timestamp:\t%i\n", (int)time(NULL));
 	compute_content_size(manifest);
 	fprintf(out, "contentsize:\t%llu\n", (long long unsigned int)manifest->contentsize);
 	includes = manifest->includes;


phmccarty:
Hmm, perhaps we could enable support for [SOURCE_DATE_EPOCH](https://reproducible-builds.org/specs/source-date-epoch/) instead? If that variable is set, we could assign the value to be the timestamps for every manifest.

matthewrsj:
but what is the point? we never use that field.

phmccarty:
The point is that a sysadmin can dig up manifests from an older build and see when those manifests were created -- timestamps embedded within a manifest are more trustworthy than their filesystem timestamps. It's for reference only, not something swupd-client would use.

matthewrsj:
ok, that seems reasonable.

phmccarty:
There was some offline discussion yesterday about the pros/cons of reproducible builds for swupd-server, and now I'm not convinced supporting SOURCE_DATE_EPOCH for manifests would be that valuable after all. At the same time, the timestamp field can be seen as valuable.

Often, if a DevOps team needs to go back to recreate a build, the purpose is to correct errors in the build, not to ensure that a bit-for-bit identical build can be produced. To validate that errors in fact were corrected in the later build, they could verify that the new manifests (and fullfile tars, etc) fixed the appropriate problems, that the manifests have newer timestamps, and are otherwise identical. With respect to reproducible builds, the team could check for identical content *except* for the timestamp fields; the timestamp field can provide better assurance that a new build was produced, and was reproducible, instead of bits from the original build being pulled in accidentally.

matthewrsj:
Ok, so builds may not be 100% reproducible due to the changed timestamps, but since the goal is not bit-by-bit reproducibility, but rather correct errors in the builds, the timestamps can actually serve as an indicator that everything else was created the same way (if the timestamp changed, but nothing else did).

I can close this unless you want to have more discussion about it.

phmccarty:
> Ok, so builds may not be 100% reproducible due to the changed timestamps, but since the goal is not bit-by-bit reproducibility, but rather correct errors in the builds, the timestamps can actually serve as an indicator that everything else was created the same way (if the timestamp changed, but nothing else did).

Yes, I agree with the above.

> I can close this unless you want to have more discussion about it.

Feel free to close :-)

PR #91, opened 2017-10-17
Enable rename support for Travis testing

phmccarty:
Because renames are an opt-in feature (i.e. not enabled by default),
pass the required configure option for Travis builds, which enables the
rename tests.


matthewrsj:
+1

PR #97, opened 2017-11-14
Enable ghosted file generation

### LINKED ISSUE #94: Orphaned renames in manifests after pruning
When rename detection is enabled, the rename detection step occurs before the pruning step, but the pruning step has no knowledge of renamed files. This can result in renamed-to files being orphaned.

To resolve this, renamed-to files should be unmarked as renamed (i.e. lose their `r` flag) in the event their corresponding renamed-from files are pruned.

matthewrsj:
Ghost files that will be deleted by third-party software on the
client-side instead of marking them as deleted in the manifest. This
allows the client to treat these as deleted when doing rename detection.
Remove those ghosted files from the manifest in the next update.

Fixes #94

### DIFF
--- a/include/swupd.h
+++ b/include/swupd.h
@@ -117,6 +117,10 @@ struct file {
 	unsigned int is_file : 1;
 	unsigned int is_link : 1;
 	unsigned int is_deleted : 1;
+	/* a ghosted file is treated as deleted so that it can be used as a rename,
+	 * but is not actually deleted in the client due to a 3rd-party program
+	 * cleaning it up. This happens with boot files managed by a boot manager */
+	unsigned int is_ghosted : 1;
 	unsigned int is_manifest : 1;
 
 	/* and these are modifiers */
@@ -194,7 +198,7 @@ extern void sort_manifest_by_version(struct manifest *manifest);
 extern bool manifest_includes(struct manifest *manifest, char *component);
 extern bool changed_includes(struct manifest *old, struct manifest *new);
 extern int prune_manifest(struct manifest *manifest);
-extern int remove_old_deleted_files(struct manifest *m1, struct manifest *m2);
+extern int remove_deprecated_files(struct manifest *m1, struct manifest *m2, bool (*compfunc)(struct file *, struct file *));
 extern void create_manifest_delta(int oldversion, int newversion, char *module);
 extern void create_manifest_deltas(struct manifest *manifest, GList *last_versions_list);
 extern void subtract_manifests_frontend(struct manifest *m1, struct manifest *m2);
--- a/src/analyze_fs.c
+++ b/src/analyze_fs.c
@@ -151,7 +151,13 @@ void populate_file_struct(struct file *file, char *filename)
 	ret = lstat(filename, &stat);
 	if (ret < 0) {
 		LOG(NULL, "stat error ", "%s: %s", filename, strerror(errno));
-		file->is_deleted = 1;
+		/* delete the file if it isn't found and isn't a boot file,
+		 * mark as ghosted if this is a boot file */
+		if (file->is_boot) {
+			file->is_ghosted = 1;
+		} else {
+			file->is_deleted = 1;
+		}
 		return;
 	}
 	file->stat.st_mode = stat.st_mode;
--- a/src/create_update.c
+++ b/src/create_update.c
@@ -236,6 +236,16 @@ static int check_group_file(void)
 	return ret;
 }
 
+static bool both_ghosted(struct file *file1, struct file *file2)
+{
+	return (file1->is_ghosted && file2->is_ghosted);
+}
+
+static bool both_deleted(struct file *file1, struct file *file2)
+{
+	return (file1->is_deleted && file2->is_deleted);
+}
+
 int main(int argc, char **argv)
 {
 	struct manifest *new_core = NULL;
@@ -252,6 +262,7 @@ int main(int argc, char **argv)
 	GList *manifests_last_versions_list = NULL;
 	int newfiles = 0;
 	int old_deleted = 0;
+	int old_ghosted = 0;
 
 	struct timeval current_time;
 	struct timeval previous_time;
@@ -325,10 +336,16 @@ int main(int argc, char **argv)
 	apply_heuristics(new_full);
 	match_manifests(old_full, new_full);
 
-	old_deleted = remove_old_deleted_files(old_full, new_full);
+	old_deleted = remove_deprecated_files(old_full, new_full, both_deleted);
 	if (old_deleted > 0) {
 		LOG(NULL, "", "Old deleted files (%d) removed from full manifest", old_deleted);
 		printf("Old deleted files (%d) removed from full manifest\n", old_deleted);
+	}
+
+	old_ghosted = remove_deprecated_files(old_full, new_full, both_ghosted);
+	if (old_ghosted > 0) {
+		LOG(NULL, "", "Old ghosted files (%d) removed from full manifest", old_ghosted);
+		printf("Old ghosted files (%d) removed from full manifest", old_ghosted);
 	}
 
 	apply_heuristics(new_full);
@@ -385,7 +402,8 @@ int main(int argc, char **argv)
 		/* Detect renamed files specifically for os-core */
 		rename_detection(new_core);
 #endif
-		old_deleted = remove_old_deleted_files(old_core, new_core);
+		old_deleted = remove_deprecated_files(old_core, new_core, both_deleted);
+		old_ghosted = remove_deprecated_files(old_core, new_core, both_ghosted);
 		sort_manifest_by_version(new_core); /* sorts by filename */
 		newfiles = prune_manifest(new_core);
 		if (newfiles <= 0) {
@@ -393,8 +411,8 @@ int main(int argc, char **argv)
 			printf("Core component has not changed (after pruning), exiting\n");
 			goto exit;
 		}
-		LOG(NULL, "", "Core component has changes (%d new, %d deleted), writing out new manifest", newfiles, old_deleted);
-		printf("Core component has changes (%d new, %d deleted), writing out new manifest\n", newfiles, old_deleted);
+		LOG(NULL, "", "Core component has changes (%d new, %d deleted, %d ghosted), writing out new manifest", newfiles, old_deleted, old_ghosted);
+		printf("Core component has changes (%d new, %d deleted, %d ghosted), writing out new manifest\n", newfiles, old_deleted, old_ghosted);
 		if (write_manifest(new_core) != 0) {
 			LOG(NULL, "", "Core component manifest write failed");
 			printf("Core component manifest write failed\n");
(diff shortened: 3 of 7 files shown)


include/swupd.h
```diff
@@ -117,6 +117,7 @@ struct file {
 	unsigned int is_file : 1;
 	unsigned int is_link : 1;
 	unsigned int is_deleted : 1;
+	unsigned int is_ghosted : 1;
```
busykai: would this be the right place to document what a "ghost" file means? otherwise one has to deduce the semantics of "ghosting" all over the place. since this changeset introduces the concept and unlike, say `is_file` it is not immediately clear, it does ask for some extra documentation (sentence or two).

src/analyze_fs.c
```diff
@@ -151,7 +151,13 @@ void populate_file_struct(struct file *file, char *filename)
 	ret = lstat(filename, &stat);
 	if (ret < 0) {
 		LOG(NULL, "stat error ", "%s: %s", filename, strerror(errno));
-		file->is_deleted = 1;
+		/* delete the file if it isn't found and isn't a boot file,
+		 * mark as ghosted if this is a boot file */
+		if (file->is_boot) {
```
bwarden, line 156: Should we be considering the other reasons lstat() might fail?

src/create_update.c
```diff
@@ -325,12 +326,18 @@ int main(int argc, char **argv)
...
+	old_deleted = remove_deprecated_files(old_full, new_full, comp_deleted);
 	if (old_deleted > 0) {
 		LOG(NULL, "", "Old deleted files (%d) removed from full manifest", old_deleted);
 		printf("Old deleted files (%d) removed from full manifest\n", old_deleted);
 	}
 
+	old_ghosted = remove_deprecated_files(old_full, new_full, comp_ghosted);
+	if (old_ghosted > 0) {
+		LOG(NULL, "", "Old ghosted files (%d) removed from full manifest", old_ghosted);
+		printf("Old deleted files (%d) removed from full manifest", old_ghosted);
```
bwarden: "Old deleted files" copy/paste holdover in printf... but could we combine these two stanzas (old_ghosted/old_deleted) logically instead?

src/create_update.c
```diff
@@ -385,16 +392,17 @@ int main(int argc, char **argv)
...
 		sort_manifest_by_version(new_core); /* sorts by filename */
 		newfiles = prune_manifest(new_core);
 		if (newfiles <= 0) {
 			LOG(NULL, "", "Core component has not changed (after pruning), exiting");
 			printf("Core component has not changed (after pruning), exiting\n");
 			goto exit;
 		}
-		LOG(NULL, "", "Core component has changes (%d new, %d deleted), writing out new manifest", newfiles, old_deleted);
-		printf("Core component has changes (%d new, %d deleted), writing out new manifest\n", newfiles, old_deleted);
+		LOG(NULL, "", "Core component has changes (%d new, %d deleted, %d ghosted), writing out new manifest", newfiles, old_deleted, old_ghosted);
```
bwarden, line 414: Could we generate one string and pass it to both LOG() and printf()?

bwarden: ...or write a single function/macro that calls both?

src/create_update.c
```diff
@@ -506,13 +514,14 @@ int main(int argc, char **argv)
...
+			old_deleted = remove_deprecated_files(oldm, newm, comp_deleted);
+			old_ghosted = remove_deprecated_files(oldm, newm, comp_ghosted);
 			sort_manifest_by_version(newm);
 			type_change_detection(newm);
 			newfiles = prune_manifest(newm);
-			if (newfiles > 0 || old_deleted > 0 || changed_includes(oldm, newm)) {
-				LOG(NULL, "", "%s component has changes (%d new, %d deleted), writing out new manifest", group, newfiles, old_deleted);
-				printf("%s component has changes (%d new, %d deleted), writing out new manifest\n", group, newfiles, old_deleted);
+			if (newfiles > 0 || old_deleted > 0 || old_ghosted > 0 || changed_includes(oldm, newm)) {
+				LOG(NULL, "", "%s component has changes (%d new, %d deleted, %d ghosted), writing out new manifest", group, newfiles, old_deleted, old_ghosted);
```
bwarden, line 533: Same about duplicate logging/printing

src/fullfiles.c
```diff
@@ -298,6 +298,11 @@ static void submit_fullfile_tasks(GList *files)
 		file = item->data;
 		item = g_list_next(item);
 
+		/* do not push ghosted files */
```
bwarden, line 301: Is this the real meat of differentiating deleted vs. ghosted files? May be the best place for detailed documentation.

src/manifest.c
```diff
@@ -242,6 +242,8 @@ struct manifest *manifest_from_file(int version, char *component)
 
 		if (c[1] == 'd') {
 			file->is_deleted = 1;
+		} else if (c[1] == 'g') {
```
bwarden: Might be time for a switch statement

src/manifest.c
```diff
@@ -428,13 +430,25 @@ int match_manifests(struct manifest *m1, struct manifest *m2)
 			file3->is_config = file1->is_config;
 			file3->is_state = file1->is_state;
 			file3->is_boot = file1->is_boot;
+			/* ghost deleted boot files */
+			file3->is_ghosted = file1->is_boot;
```
bwarden: Should this be `= file1->is_boot && file1->is_deleted`?

src/manifest.c
```diff
@@ -428,13 +430,25 @@ int match_manifests(struct manifest *m1, struct manifest *m2)
...
-			if (!file1->is_deleted) {
+			if (file3->is_ghosted) {
+				/* if the new file is ghosted, preserve the hash and rename
+				 * information and set the last_change version to the current
+				 * version. */
 				file3->last_change = m2->version;
-			} else {
+				file3->is_rename = file1->is_rename;
+				hash_assign(file1->hash, file3->hash);
+			} else if (file1->is_deleted) {
```
bwarden: Only `file3->last_change` is different between these two conditions; consider condensing them into one block, with a smaller conditional (or trinary) for `file3->last_change` only.

bwarden: Or we even use `m2->version` for everything else... I think this can be condensed a lot, into single-line conditionals.

src/manifest.c
```diff
@@ -658,6 +672,10 @@ char *file_type_to_string(struct file *file)
 		type[1] = 'd';
 	}
 
+	if (file->is_ghosted) {
```
bwarden, line 678: Switch statement instead?

include/swupd.h
matthewrsj: Sure, it doesn't hurt to add a comment here if that's what you mean.

src/manifest.c
```diff
@@ -968,11 +986,12 @@ bool changed_includes(struct manifest *old, struct manifest *new)
  *
  * Note: this function should be called after match_manifests().
  */
-int remove_old_deleted_files(struct manifest *m1, struct manifest *m2)
+int remove_deprecated_files(struct manifest *m1, struct manifest *m2, void *compfunc)
 {
 	GList *list1, *list2;
 	struct file *file1, *file2;
 	int count = 0;
+	bool (*compare)(struct file *, struct file *) = compfunc;
```
bwarden: Better check for `compfunc=NULL` before you call it below. Also, would an `int` be better return type for this so you can report/handle errors from a comparison function, like when `file=NULL`?

src/helpers.c
```diff
@@ -324,3 +324,13 @@ bool file_is_debuginfo(const char *path)
 	free(src);
 	return ret;
 }
+
+bool comp_ghosted(struct file *file1, struct file *file2)
+{
+	return (file1->is_ghosted && file2->is_ghosted);
```
bwarden: Really should check if `file1` or `file2` is null.

bwarden (commented):
There are a few copy/paste sections where, although I only found one error, it's easy for maintainers to make subsequent errors. I'd rather see duplicated code combined, and only the specific individual statements that are different be surrounded by conditionals.

src/analyze_fs.c
icarus-sparry: In an ideal world, then yes. However of the 8 error returns for lstat, 
EACCESS: Should not get as we are running as root
EFAULT: Should not get as we are passing a valid address for &stat
ELOOP: Should not get, we are constructing the file tree.
ENAMETOOLONG: Should not get as we are constructing the file tree
ENOMEM: If we have run out of kernel memory we are in big trouble.
ENOTDIR: Should not get as we are constructing the file tree
EOVERFLOF: If we are running on a 32 bit machine and have more than 4GB files, then yes, but otherwise no.

Which just leaves ENOENT, which is the case that is being handled. I suppose we could add an abort() if we don't get ENOENT.

src/create_update.c
icarus-sparry: It would be nice to have the single function.

src/manifest.c
icarus-sparry: +1

include/swupd.h
```diff
@@ -194,7 +195,7 @@ extern void sort_manifest_by_version(struct manifest *manifest);
 extern bool manifest_includes(struct manifest *manifest, char *component);
 extern bool changed_includes(struct manifest *old, struct manifest *new);
 extern int prune_manifest(struct manifest *manifest);
-extern int remove_old_deleted_files(struct manifest *m1, struct manifest *m2);
+extern int remove_deprecated_files(struct manifest *m1, struct manifest *m2, void *compfunc);
```
icarus-sparry: We should declare compfunc correctly

extern int remove_deprecated_files(struct manifest *m1, struct manifest *m2, bool(*compfunc)(struct file *,struct file*));

include/swupd.h
```diff
@@ -275,5 +276,7 @@ extern void move_fd(int oldfd, int newfd);
 extern int wait_process_terminate(pid_t pid);
 extern int num_threads(float scaling);
 extern bool file_is_debuginfo(const char *path);
+extern bool comp_deleted(struct file *file1, struct file *file2);
```
icarus-sparry: Not sure why these internal functions are being declared external here.

src/helpers.c
icarus-sparry: Since this is only used in create_update.c it should go in there and not be declared external.

I don't think it should check for file1 and file2 being not null, it is not clear what the return value should be in that case.

Same comment on the name as for comp_deleted

src/helpers.c
```diff
@@ -324,3 +324,13 @@ bool file_is_debuginfo(const char *path)
...
+}
+
+bool comp_deleted(struct file *file1, struct file *file2)
```
icarus-sparry: Also should move.

Maybe rename to both_deleted, I would expect from the name that the code would be
return (file1->is_deleted) == (file2->is_deleted);

src/manifest.c
```diff
@@ -968,11 +986,12 @@ bool changed_includes(struct manifest *old, struct manifest *new)
  *
  * Note: this function should be called after match_manifests().
  */
-int remove_old_deleted_files(struct manifest *m1, struct manifest *m2)
+int remove_deprecated_files(struct manifest *m1, struct manifest *m2, void *compfunc)
```
icarus-sparry: Just declare it properly in the first place,then you don't have to cast it later.

src/manifest.c
icarus-sparry: I think it is reasonable to say the thing calling the comparison function should know if the pointer is Null or not. cf strcmp. So in that case bool is fine as a return type.

src/create_update.c
busykai: while i agree, it looks ugly. i'd rather file a separate issue, pointing out examples. because, firstly, i doubt changes in this PR affected all occurrences of duplicating output; secondly, it's our of the scope of this particular change to modify the behavior.

src/helpers.c
busykai: +1 on moving (and removing declarations from `swupd.h`) and not checking for `NULL` (the caller must ensure that).

src/analyze_fs.c
matthewrsj: Regardless, adding those checks is out of the scope of this PR.

src/create_update.c
matthewrsj: I agree it would be nice to have a single function, but as @busykai pointed out  the duplicate `LOG` and `printf` calls happen elsewhere in the code as well. They should all be fixed up together.

src/fullfiles.c
matthewrsj: No, not really. This is just making sure we aren't creating fullfiles for ghosted files. Ghosted files are much closer to deleted files than fullfiles.

src/manifest.c
matthewrsj: If you look up a few lines you'll see that file1 is deleted if we get to this point, but it is nice to add the extra check here in case our logic fails (although that would never ever happen of course /s).

src/manifest.c
matthewrsj: Doesn't work as well going this direction (checking a bunch of different `int` fields and assigning `char`s). Anyways that's a restructure of the function and I'm just adding this single conditional so I'd call that out-of-scope for this change.

matthewrsj:
Comments addressed.

include/swupd.h
```diff
@@ -117,6 +117,10 @@ struct file {
 	unsigned int is_file : 1;
 	unsigned int is_link : 1;
 	unsigned int is_deleted : 1;
+	/* a ghosted file is treated as deleted so that it can be used as a rename,
+	 * but is not actually deleted in the client due to a 3rd-parth program
```
icarus-sparry: spelling, parth->party

icarus-sparry (commented):
lgtm

busykai:
+1

P.S. "parth" is a legit word in Welsh, I read... :)

PR #98, opened 2017-11-14
Prune orphaned rename files from manifest

### LINKED ISSUE #93: Lingering renames in manifests after target deleted
In the case where renames are detected and included in a manifest (i.e. version 10) for a bundle, then the target file is later deleted from that bundle, the renamed-from file will hang around with no corresponding target.

These renamed-from files (`.d.r`) should have their hashes set to all zeros and turned into a pure delete when the matching hash does not exist in the manifest.

matthewrsj:
When renamed-to files are deleted or are renamed again, the original
renamed-from file is orphaned as it has no renamed-to peer. Prune these
from the manifests by checking if the rename_peer field is NULL.

Fixes #93

### DIFF
--- a/test/functional/orphaned-renames/test.bats
+++ b/test/functional/orphaned-renames/test.bats
@@ -1,0 +1,61 @@
+#!/usr/bin/env bats
+
+# common functions
+load "../swupdlib"
+
+setup() {
+  clean_test_dir
+  init_test_dir
+
+  init_server_ini
+  set_latest_ver 0
+  init_groups_ini os-core test-bundle
+
+  set_os_release 10 os-core
+  set_os_release 10 test-bundle
+  track_bundle 10 os-core
+  track_bundle 10 test-bundle
+
+  set_os_release 20 os-core
+  set_os_release 20 test-bundle
+  track_bundle 20 os-core
+  track_bundle 20 test-bundle
+
+  set_os_release 30 os-core
+  set_os_release 30 test-bundle
+  track_bundle 30 os-core
+  track_bundle 30 test-bundle
+
+  gen_file_plain_with_content 10 test-bundle /usr/lib/bar "$(seq 100)"
+  gen_file_plain_with_content 20 test-bundle /usr/lib/baz "$(seq 100)"
+  # different content just to make sure this works with delta renames as well as
+  # direct renames
+  gen_file_plain_with_content 30 test-bundle /usr/lib/foo "$(seq 100) new"
+}
+
+@test "create updates with renamed-to file getting deleted" {
+  sudo $CREATE_UPDATE --osversion 10 --statedir $DIR --format 3
+  sudo $MAKE_FULLFILES --statedir $DIR 10
+
+  set_latest_ver 10
+
+  sudo $CREATE_UPDATE --osversion 20 --statedir $DIR --format 3
+  sudo $MAKE_FULLFILES --statedir $DIR 20
+
+  set_latest_ver 20
+
+  sudo $CREATE_UPDATE --osversion 30 --statedir $DIR --format 3
+  sudo $MAKE_FULLFILES --statedir $DIR 30
+  # version 10: add file to 10
+  [ 1 -eq $(grep 'F\.\.\.	.*	10	/usr/lib/bar' $DIR/www/10/Manifest.test-bundle | wc -l) ]
+  # version 20: rename bar to baz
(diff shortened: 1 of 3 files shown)


test/functional/orphaned-renames/test.bats
```diff
@@ -0,0 +1,61 @@
...
+  sudo $CREATE_UPDATE --osversion 30 --statedir $DIR --format 3
+  sudo $MAKE_FULLFILES --statedir $DIR 30
+  # version 10: add file to 10
+  [ 1 -eq $(grep 'F...	.*	10	/usr/lib/bar' $DIR/www/10/Manifest.test-bundle | wc -l) ]
+  # version 20: rename bar to baz
+  [ 1 -eq $(grep '\.d\.r	.*	20	/usr/lib/bar' $DIR/www/20/Manifest.test-bundle | wc -l) ]
+  [ 1 -eq $(grep 'F\.\.r	.*	20	/usr/lib/baz' $DIR/www/20/Manifest.test-bundle | wc -l) ]
+  # version 30: delete renamed-to file (baz), should clean up renamed-from file (baz)
+  # Check for the new renamed-to file (foo)
+  [ 0 -eq $(grep '/usr/lib/bar' $DIR/www/30/Manifest.test-bundle | wc -l) ]
```
busykai: w/o this fix, is this the line which would fail?

busykai (commented):
Looks good to me. I had one question on the tests.

I'm also curious how it would behave in case `bar` and `baz` present in version 10 and only `foo` present in version 20. In other words how a peer is determined where there are more than one candidates. Would this be out of scope for this change?

test/functional/orphaned-renames/test.bats
icarus-sparry: If I understand the question, it is not well determined. There are a number of heuristics which go into a scoring algorithm, and the file with the best score wins. If there is a tie then you are at the mercy of the sort algorithm as to which is chosen.

matthewrsj: Any of these things could fail if the cleanup is done incorrectly. Notice I'm checking lines in the manifests for versions 10, 20, and 30.

First, I expect /usr/lib/bar to be deleted but a rename in 20, with baz as the target. Then in 30 bar must be cleaned up because baz is now renamed to foo so is deleted itself.

busykai: I understand the logic of the test. I guess I did not quite understood if current implementation would generate a different manifest.

The change prunes orphans from the manifest. So who's the orphan we're catching with the tests? I assumed it would be `/usr/lib/bar` and hence existing implementation would fail since it would generate the manifest with `/usr/lib/bar` as `.d.r` in it. If current implementation produces the same manifest, then the question is how the test tests the change.

test/functional/orphaned-renames/test.bats
```diff
@@ -0,0 +1,61 @@
...
+  set_latest_ver 20
+
+  sudo $CREATE_UPDATE --osversion 30 --statedir $DIR --format 3
+  sudo $MAKE_FULLFILES --statedir $DIR 30
+  # version 10: add file to 10
+  [ 1 -eq $(grep 'F...	.*	10	/usr/lib/bar' $DIR/www/10/Manifest.test-bundle | wc -l) ]
+  # version 20: rename bar to baz
+  [ 1 -eq $(grep '\.d\.r	.*	20	/usr/lib/bar' $DIR/www/20/Manifest.test-bundle | wc -l) ]
+  [ 1 -eq $(grep 'F\.\.r	.*	20	/usr/lib/baz' $DIR/www/20/Manifest.test-bundle | wc -l) ]
+  # version 30: delete renamed-to file (baz), should clean up renamed-from file (baz)
```
busykai: is this comment correct? in version 30 baz is rename-from and it should be deleted.

matthewrsj: eh, this comment is so not correct

test/functional/orphaned-renames/test.bats
matthewrsj: We are doing two updates 10 -> 20 and 20 -> 30. In 10 -> 20 we are renaming bar to baz, everything will be the same as the current implementation here, no pruning is needed. When we do the update from 20 -> 30, we are renaming the file again (baz -> foo). At this point bar (original file) is orphaned and would not be removed by current implementation, but is removed by the implementation in this pr.

The tests for the update from 10 -> 20 are just to verify that the tests for 20 -> 30 mean something (if bar wasn't in 10 or 20 it of course wouldn't be in 30 unless it is added in the test) and that the flags are correct.

tmarcu:
lgtm

PR #101, opened 2017-11-27
Do not delete renames from old versions

matthewrsj:
Perform a final link on all renames in the manifest in order to track
renames from older versions. Do not remove orphaned renames but instead
keep them around as deleted files as well.

### DIFF
--- a/src/rename.c
+++ b/src/rename.c
@@ -446,6 +446,48 @@ redo:
 	g_list_free(deleted_files);
 }
 
+void final_link(GList *files)
+{
+	GList *list1, *list2;
+	struct file *file1, *file2;
+
+	list1 = new_list_renamed_files(files);
+	list1 = list2 = g_list_sort(list1, file_sort_version);
+
+	for (; list1; list1 = g_list_next(list1)) {
+		file1 = list1->data;
+		/* do not re-link already-linked renames */
+		if (file1->rename_peer != NULL) {
+			continue;
+		}
+
+		/* check the rest of the list */
+		list2 = g_list_next(list1);
+		for (; list2; list2 = g_list_next(list2)) {
+			file2 = list2->data;
+			/* must be the same version to link */
+			if (file1->last_change != file2->last_change) {
+				continue;
+			}
+
+			/* need one and only one to be deleted to link as this
+			 * indicates a renamed-from/renamed-to relationship
+			 * (.d.r -> F..r) */
+			if (file1->is_deleted == file2->is_deleted) {
+				continue;
+			}
+
+			/* must have same hash to link */
+			if (!hash_compare(file2->hash, file1->hash)) {
+				continue;
+			}
+
+			file1->rename_peer = file2;
+			file2->rename_peer = file1;
+		}
+	}
+}
+
 /* What do we need this for?
  *
  * rename_detection has already set up the links in the manifest it
--- a/test/functional/orphaned-renames/test.bats
+++ b/test/functional/orphaned-renames/test.bats
@@ -26,11 +26,25 @@ setup() {
   track_bundle 30 os-core
   track_bundle 30 test-bundle
 
+  set_os_release 40 os-core
+  set_os_release 40 test-bundle
+  track_bundle 40 os-core
+  track_bundle 40 test-bundle
+
+  # /usr/lib/bar and /one will be renamed to /usr/lib/baz and /two
   gen_file_plain_with_content 10 test-bundle /usr/lib/bar "$(seq 100)"
+  gen_file_plain_with_content 10 test-bundle /one "$(printf 'a%.0s' {1..200})"
+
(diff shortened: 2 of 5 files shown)


src/rename.c
```diff
@@ -446,6 +446,43 @@ void rename_detection(struct manifest *manifest)
...
+		list2 = g_list_next(list1);
+		for (; list2; list2 = g_list_next(list2)) {
+			file2 = list2->data;
+			/* must have same hash to link */
+			if (!hash_compare(file2->hash, file1->hash)) {
+				continue;
+			}
+
+			/* must be the same version to link */
+			if (file1->last_change != file2->last_change) {
```
phmccarty: Since you are sorting by version, it would probably be better to check this condition first (before the hash_compare).

test/functional/orphaned-renames/test.bats
```diff
@@ -46,16 +60,35 @@ setup() {
...
+  [ 1 -eq $(grep $'F\.\.r\t.*\t20\t/two' $DIR/www/20/Manifest.test-bundle | wc -l) ]
+  [ 1 -eq $(grep $'\.d\.r\t.*\t20\t/usr/lib/bar' $DIR/www/20/Manifest.test-bundle | wc -l) ]
+  [ 1 -eq $(grep $'F\.\.r\t.*\t20\t/usr/lib/baz' $DIR/www/20/Manifest.test-bundle | wc -l) ]
   # version 30: prune original renamed-from file (bar), baz is now a renamed-from file
   # Check for the new renamed-to file (foo)
-  [ 0 -eq $(grep '/usr/lib/bar' $DIR/www/30/Manifest.test-bundle | wc -l) ]
-  [ 1 -eq $(grep '\.d\.r	.*	30	/usr/lib/baz' $DIR/www/30/Manifest.test-bundle | wc -l) ]
-  [ 1 -eq $(grep 'F\.\.r	.*	30	/usr/lib/foo' $DIR/www/30/Manifest.test-bundle | wc -l) ]
+  # /two was deleted in this version, so both /one and /two should be marked as deleted
+  [ 1 -eq $(grep $'\.d\.\.\t0000000000000000000000000000000000000000000000000000000000000000\t20\t/one' $DIR/www/30/Manifest.test-bundle | wc -l) ]
```
phmccarty: Not sure if other tests write out the zeros like above, but you can make the regexp shorter. For example: `\.d\.\.\t0\{64\}\t20\t/one`.

test/functional/orphaned-renames/test.bats
```diff
@@ -46,16 +60,35 @@ setup() {
...
+  sudo $MAKE_FULLFILES --statedir $DIR 40
+  # version 10: add files to 10
+  [ 1 -eq $(grep $'F\.\.\.\t.*\t10\t/usr/lib/bar' $DIR/www/10/Manifest.test-bundle | wc -l) ]
+  [ 1 -eq $(grep $'F\.\.\.\t.*\t10\t/one' $DIR/www/10/Manifest.test-bundle | wc -l) ]
+  # version 20: rename bar to baz and one to two
+  [ 1 -eq $(grep $'\.d\.r\t.*\t20\t/one' $DIR/www/20/Manifest.test-bundle | wc -l) ]
+  [ 1 -eq $(grep $'F\.\.r\t.*\t20\t/two' $DIR/www/20/Manifest.test-bundle | wc -l) ]
+  [ 1 -eq $(grep $'\.d\.r\t.*\t20\t/usr/lib/bar' $DIR/www/20/Manifest.test-bundle | wc -l) ]
+  [ 1 -eq $(grep $'F\.\.r\t.*\t20\t/usr/lib/baz' $DIR/www/20/Manifest.test-bundle | wc -l) ]
   # version 30: prune original renamed-from file (bar), baz is now a renamed-from file
```
phmccarty: This comment looks out-of-date.

src/rename.c
matthewrsj: yes, good point.

test/functional/orphaned-renames/test.bats
matthewrsj: right, ok.

src/rename.c
```diff
@@ -446,6 +446,43 @@ void rename_detection(struct manifest *manifest)
...
+
+			/* need one and only one to be deleted to link as this
+			 * indicates a renamed-from/renamed-to relationship
+			 * (.d.r -> F..r) */
+			if (file1->is_deleted == file2->is_deleted) {
+				continue;
+			}
+
+			file1->rename_peer = file2;
+			file2->rename_peer = file1;
```
phmccarty, line 486: If rename_peer is set for a file on subsequent passes through this loop, should you check for that condition and continue?

matthewrsj: yes, we should check file1->rename_peer before descending into the second loop.

matthewrsj:
Comments addressed @phmccarty

phmccarty (approved):
+1, looks good

PR #102, opened 2017-11-28
Fix ghosted-file deprecation

matthewrsj:
Ghosted files were only being deprecated over format bumps. Move the
format-bump-check logic out to the calling function to allow ghosted
files to be removed on every build. Also assign the is_ghosted flag when
adding "deleted" files from the old manifest so that the file will
actually be removed.


phmccarty (approved):
+1

PR #103, opened 2017-11-30
Refactor match_manifests function

matthewrsj:
This function had multiple sections of copy-pasted code. This was
causing an issue of bug-fixes being added to one part of code but not
the other. This patch refactors it to use internal functions where
appropriate and removes some dead code.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -323,6 +323,21 @@ struct manifest *manifest_from_file(int version, char *component)
 	return manifest;
 }
 
+/* do not set peer for the case where the old file was deleted and the new file
+ * was added back. This peer is normally set when the names match up and this
+ * triggers an attempt at creating a delta during pack creation. For this case
+ * where the old file is deleted or ghosted and is brought back we do not want
+ * this to be attempted, as the old file will not be present when trying to
+ * create the delta.
+ *
+ * We can go even further and not set peers for files when the old one did not
+ * exist (deleted or ghosted). It is not necessary to check the status of the
+ * new file */
+bool should_have_peer(struct file *file1, struct file *file2)
+{
+	return (!(file1->is_deleted || file1->is_ghosted));
+}
+
 void free_manifest(struct manifest *manifest)
 {
 	struct file *file;
@@ -340,6 +355,62 @@ void free_manifest(struct manifest *manifest)
 	free(manifest);
 }
 
+static bool same_file_contents(struct file *file1, struct file *file2)
+{
+	return (file1->is_dir == file2->is_dir &&
+		file1->is_link == file2->is_link &&
+		file1->is_deleted == file2->is_deleted &&
+		file1->is_file == file2->is_file &&
+		file1->is_config == file2->is_config &&
+		file1->is_state == file2->is_state &&
+		file1->is_boot == file2->is_boot &&
+		hash_compare(file1->hash, file2->hash));
+}
+
+/*
+ * Add a deleted file entry for it in the target list. However, since we're
+ * currently walking the list we HAVE to prepend the entry. Calling function
+ * should track to sort at the end.
+ */
+static void add_deleted_file(struct file *source, struct manifest *manifest)
+{
+	struct file *deleted;
+	deleted = calloc(1, sizeof(struct file));
+	if (deleted == NULL) {
+		assert(0);
+	}
+
+	deleted->filename = strdup(source->filename);
+	hash_set_zeros(deleted->hash);
+	deleted->is_deleted = 1;
+	deleted->is_config = source->is_config;
+	deleted->is_state = source->is_state;
+	deleted->is_boot = source->is_boot;
+	/* ghost deleted boot files */
+	deleted->is_ghosted = source->is_ghosted | source->is_boot & source->is_deleted;
+
+	if (deleted->is_ghosted || source->is_deleted) {
+		/* if the new file is ghosted or the file was deleted, preserve
+		 * hash (all zeros if source->is_deleted) and rename status */
+		deleted->is_rename = source->is_rename;
+		hash_assign(source->hash, deleted->hash);
+	}
+
+	/* for deleted files last_change remains the same.
+	 * otherwise last change is now */
+	deleted->last_change = source->is_deleted ? source->last_change : manifest->version;
+
+	deleted->peer = source;
+	source->peer = deleted;
+
+	/* if we are adding a deleted file we are walking the old and new
+	 * manifest files in-sync. we need to prepend this file in order to
+	 * not process it twice. This is why it is important for the calling
+	 * function to sort the list again at the end */
+	manifest->files = g_list_prepend(manifest->files, deleted);
+	manifest->count++;
+}
+
 /*
 backfill the "last changed" of each file in a manifest
 by comparing the hash against the same file in the previous manifest
@@ -354,7 +425,6 @@ int match_manifests(struct manifest *m1, struct manifest *m2)
 	struct file *file1, *file2;
 	int must_sort = 0;
 	int count = 0;
-	int first = 1;
 
 	if (!m1) {
 		printf("Matching manifests up failed: No old manifest!\n");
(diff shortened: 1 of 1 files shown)


src/manifest.c
```diff
@@ -340,6 +340,61 @@ void free_manifest(struct manifest *manifest)
 	free(manifest);
 }
 
+static bool same_file(struct file *file1, struct file *file2)
+{
+	return (hash_compare(file1->hash, file2->hash) &&
```
icarus-sparry: Probably reasonable to change the order of this to do the cheap compares first, and the hash_compare last, at least until I make hash_compare cheap.

src/manifest.c
```diff
@@ -340,6 +340,61 @@ void free_manifest(struct manifest *manifest)
 	free(manifest);
 }
 
+static bool same_file(struct file *file1, struct file *file2)
```
icarus-sparry: Perhaps a better name might be 'same_file_contents', as the names might differ.

src/manifest.c
```diff
@@ -382,166 +436,81 @@ int match_manifests(struct manifest *m1, struct manifest *m2)
...
-			 * otherwise last change is now */
-			file3->last_change = file1->is_deleted ? file1->last_change : m2->version;
-
-			file3->peer = file1;
-			file1->peer = file3;
-
-			list1 = g_list_next(list1);
-			m2->files = g_list_prepend(m2->files, file3);
-			m2->count++;
+		} else if (ret < 0) {
```
icarus-sparry, line 475: I would prefer this to read

       list2 = g_list_next(list2);
       continue;
    }
    if (ret < 0) {`

as this makes clear that nothing more is being done. This is purely a readability change, I expect there to be no difference in the generated code.

src/manifest.c
```diff
@@ -382,166 +436,81 @@ int match_manifests(struct manifest *m1, struct manifest *m2)
...
-					LOG(file1, "file got deleted", "");
-					first = 0;
-				}
 			}
+
 			must_sort = 1;
-			continue;
+			/* advance list1 for next file */
+			list1 = g_list_next(list1);
+		} else {
```
icarus-sparry, line 486: I would like a `continue` in here as well.

src/manifest.c
```diff
@@ -382,166 +436,81 @@ int match_manifests(struct manifest *m1, struct manifest *m2)
...
 		}
-		/* if we get here, ret is > 0, which means this is a new file added */
-		/* all we do is advance the pointer */
-		account_new_file();
-		list2 = g_list_next(list2);
-		count++;
 	}
 
 	/* now deal with the tail ends */
 	while (list1) {
```
icarus-sparry: I would prefer this to be a for loop, rather than a while loop.

src/manifest.c
```diff
@@ -340,6 +340,61 @@ void free_manifest(struct manifest *manifest)
...
+	}
+
+	/* for deleted files last_change remains the same.
+	 * otherwise last change is now */
+	deleted->last_change = source->is_deleted ? source->last_change : manifest->version;
+
+	deleted->peer = source;
+	source->peer = deleted;
+
+	manifest->files = g_list_prepend(manifest->files, deleted);
```
icarus-sparry: I think you should copy the comment from the old lines 496 to 500, explaining you are prepending here because there may (and almost certainly is) a list traversal in progress.

If we ever get around to changing the data structures to use arrays rather than linked lists this subtle point would be easy to overlook.

src/manifest.c
```diff
@@ -382,166 +436,81 @@ int match_manifests(struct manifest *m1, struct manifest *m2)
...
-			if (!file1->is_deleted || file2->is_deleted) {
+			/* do not set peer for the case where the old file was
+			 * deleted and the new file was added back. This peer
+			 * is normally set when the names match up and this
+			 * triggers an attempt at creating a delta during pack
+			 * creation. For this case where the old file is deleted
+			 * or ghosted and is brought back we do not want this to
+			 * be attempted, as the old file will not be present
+			 * when trying to create the delta. */
+			if (!(file1->is_deleted && !file2->is_deleted) ||
```
icarus-sparry: I would like this comment and conditional moved into a tiny extra function, and given a good name, perhaps `risen_from_death(file1,file2)` where the reasons can be spelt out in detail.

    {
        if (! file1->is_deleted) return true; /* file should exist on target device unless deleted/ghosted */
        if (! file1->is_ghosted) return true;
        if (file2->is_deleted) return true; /* Normal delete */
        if (file2->is_ghosted) return true; /* ghosted delete */
        return false; /* special case of resurected file */
        }

possibly a second helper function to say file->is_ghosted || file->is_deleted ?

OK, risen_from_death is a lousy name,, "should_have_peer" maybe.

src/manifest.c
```diff
@@ -340,6 +340,61 @@ void free_manifest(struct manifest *manifest)
...
+
+	deleted->filename = strdup(source->filename);
+	hash_set_zeros(deleted->hash);
+	deleted->is_deleted = 1;
+	deleted->is_config = source->is_config;
+	deleted->is_state = source->is_state;
+	deleted->is_boot = source->is_boot;
+	deleted->is_ghosted = source->is_ghosted;
+	/* ghost deleted boot files */
+	if (!deleted->is_ghosted) {
```
icarus-sparry: Maybe

deleted->is_ghosted |= source->is_boot & source->is_deleted

icarus-sparry (commented):
A few changes to help me with comprehension of the code, but generally good.

src/manifest.c
matthewrsj: +1

src/manifest.c
matthewrsj: I put it in the function comment but yes I'll add it here as well.

src/manifest.c
matthewrsj: yes, that would be easier to understand.

src/manifest.c
matthewrsj: I disagree here. Each of the three conditionals is testing the return from a single `strcmp`, which can either be <0, 0, or >0. This seems much more natural as an `if/else if/else` to me. Having three separate `if` conditionals confuses the fact that we are checking the same thing.

src/manifest.c
matthewrsj: Why? `while (something in list)` seems intuitive enough to me.

src/manifest.c
matthewrsj: Ah, true. As far as this function is concerned they can have different filenames, even though where it is called from dictates that they will always have the same name. I'll update the function name.

src/manifest.c
icarus-sparry: My point is I read code from beginning to end. If the code was small, i.e. no more than a couple of lines in the bodies of the if, the elif and the else so I could see the entire structure of the loop on the screen I would agree. When you can not see the end of the else you have to mentally push the state and look to see what happens next, then return to the current place. With a continue statement one knows that you are finished with the processing, as you are going back to the top of the loop.

If you want to keep the if/else if/else structure, then nothing is stopping you putting in a continue, but unconditional changes of flow just before else statements looks wrong to me.

matthewrsj: The if/else if/else spans from line 438 to 483 - 45 lines. This fits on the same page of any modern monitor. The conditional will be even shorter once I incorporate your `resurrected_from_the_dead` function and take the huge comment out of the if clause.

src/manifest.c
icarus-sparry: I think there is a big difference between 'for each item in list' and 'while the list is not empty'. The latter is clearly a superset of the former.

So arguments for using for are
1. If you don't need the extra power that a while loop gives you, then you are conveying more accurately to your reader what you are wanting.
1. If we get to change these data structures to use arrays then we only need to fix the re-init in one place, rather than potentially many.
1. remember we have had bugs where people have had continue statements which lead to infinite loops because they have not moved on to the next statement.

As I said in my review, these changes are suggestions to make the code more readable, at least to me. They are just suggestions. At the end of the day it is your code.

matthewrsj: > If we get to change these data structures to use arrays then we only need to fix the re-init in one place, rather than potentially many.

this convinced me.

matthewrsj:
Comments addressed.

src/manifest.c
phmccarty: I prefer the if/else if/else form in this case, for similar reasons as @matthewrsj described in his first comment here. With this amount of refactoring, there is very little code in each block, and I think it reads okay like this.

phmccarty (commented):
Changes look fine to me. +1

PR #104, opened 2017-11-30
test: fix some typos in checkrenamed()

phmccarty:
Due to these typos being present, several tests were passing that should not have been.

This is not merge-ready yet; I'm opening it to track the resultant test failures and work to fix them.

### DIFF
--- a/test/functional/renames/test.bats
+++ b/test/functional/renames/test.bats
@@ -47,7 +47,7 @@ gendataCs() {
 }
 
 checkrenamed(){
-    local flags sh1 ver name fromsha1="bad" tosha1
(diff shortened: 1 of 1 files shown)


matthewrsj:
Interesting that test-suite.log contains all "passed" tests as well.  Makes it a little harder to glance and see what the errors are.

phmccarty:
@matthewrsj Yes, the log contains info on passing tests, but their stderr/stdout output is not printed, so the vast majority of the log contents are related to the failing tests.

matthewrsj:
Although this project is no longer under active development, I am merging this PR so that anyone picking up the project is aware of these test failures, as they should be addressed by anyone using this codebase.

PR #105, opened 2017-12-01
 Do not attempt full file creation for ghosted files

matthewrsj:
Swupd logs showed that fullfile creation was being attempted for ghosted files. This should not happen.


matthewrsj:
oops, this isn't right.

PR #106, opened 2017-12-01
Do not attempt full-file creation for ghosted files

matthewrsj:
Ghosted files should obviously not have full-files created, but swupd build logs showed that swupd was attempting to create them.


phmccarty (commented):
Changes look good to me. +1

PR #109, opened 2017-12-06
heuristics: set /usr/src/kernel as not state dir

### LINKED ISSUE #108: Add /usr/src/kernel as a no state dir
Clear Linux will ship the Linux kernel source to `/usr/src/kernel` directory.

Currently this is a state directory and the content wont be shipped.

miguelinux:
The kernel sources will be shipped in /usr/src/kernel but /usr/src/* is
state.

This commit enables /usr/src/kernel and its contents as a no state
directory.

Fixes #108

Signed-off-by: Miguel Bernal Marin <miguel.bernal.marin@linux.intel.com>


matthewrsj:
I don't love introducing *more* specific use cases but meh, this function isn't pretty already. +1

